PR #94, opened 2013-06-06
Adding the ability to download a tinderbox build based on the revision passed in

AutomatedTester:
I need to be able to download builds based on a revision as part of Bisect in the cloud so this gives me what I want but am happy to clean up as needed


whimboo:
Closing out the branch given no activity for nearly 2 years.

Issue #211, opened 2014-05-22
Change our current progressbar package


Nebelhom:
Our current progressbar library is not maintained any more (last update 2011), has some shortcomings with regards to unknown file sizes (see #194 ) and is not easily extendible.

It might be best to switch to a new library here and while we're at it, maybe take a lightweight one that is easily extendible and fit it to our needs.

Whimboo has already made a suggestion:

https://pypi.python.org/pypi/progress/1.2

Last update was Nov 2013, it looks fairly lightweight and for our most current issue, it has the option of using a spinner instead of a progressbar, if we do not know the file size.

Are there any other suggestions? We could always start from scratch... but I, at least, realistically cannot guarantee that I will have consistently time to write a completely new package. If anyone else fancies that, feel free to comment.

Nebelhom:
@whimboo @davehunt 

FYI.

Looks like someone picked up the old progressbar package, forked it from the source and carried it on.

https://pypi.python.org/pypi/progressbar2

whimboo:
Yeah, I noticed that again today after forgetting about your comment. Looks like we have another chance! Maybe lets see what's new and if it helps us.

whimboo:
Nice. It may be good to check that out. Labeling issue as mentored in case someone has interest to look at this.

whimboo:
There is also https://github.com/tqdm/tqdm, but maybe we should just change to progressbar2, and upgrade to its latest release. It's still under active development.

nathanielnebel:
I would be interested in working on this bug, would it still be possible to receive mentoring?

whimboo:
Hello @nathanielnebel. Sure, this issue is still available to pick up. And I'm happy to mentor.

Given the still open question of what to use, I would say we keep progressbar and update it to progressbar2. There should be minimal efforts necessary to do that, and would be a good starting point for you.

Let me know what kind of issues you have or if something is not clear. I will respond as soon as I can. Thanks!

nathanielnebel:
Changes are in pull request# 452.

whimboo:
Hi @nathanielnebel. Thank you for doing the work, and it's good to hear that we don't need any changes in our code. I will have a look at your PR in a moment.

whimboo:
PR #453 got merged now.

whimboo:
@nathanielnebel in case you are interested to contribute a bit more please let me know and we can certainly find a new issue for you, or maybe you can find yourself from the list of open ones.

[...]

PR #213, opened 2014-05-24
Exchange progressbar with progress (#211)

Nebelhom:
This PR addresses issue #211

I thought, I'd kick this off with a first implementation of using progress. The change was easy enough and it is easy to handle.

It still needs a test for a website with no content-header and I would need a site with a larger downloadable file without a content-header to see how the spinner works.

I know it still needs work, but I would appreciate some input, if we should go ahead with it.

Cheers

### DIFF
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -22,7 +22,8 @@ from parser import DirectoryParser
 from timezones import PacificTimezone
 from utils import urljoin
 
-import progressbar as pb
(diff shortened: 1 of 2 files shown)


Nebelhom:
@whimboo 

I accidentally closed this during clean up, but I will leave it closed as it has gone stale.

It is not a priority and once it becomes one again, I would rather start from fresh than re-use a bitrotted PR.

Issue #22, opened 2012-08-17
Add a bash-completion script


aleksejrs:
mozdownload has multiple command line switches, some of which are only available in the long form.
mozdownload also has 43-line help output.

bash-completion allows for auto-completion of command line options of a program when using bash or zsh.  For example, git provides a 56-kilobyte description of completion in Debian (most auto-completion scripts are only a couple kilobytes though).  With it, you can enter “mozdownload --”, press Tab, and see a concise list of all the available long switches; or you can enter also a couple of letters of a switch and press Tab and have that switch completed.

Most of the auto-completion scripts are written in bash itself, but there is a Perl support script, and at least one program to generate one based on Python's optparse syntax (possibly supporting or ported to argparse).

whimboo:
An implementation here would be blocked by issue #269.

PR #222, opened 2014-06-27
Add a bash-completion script (#22)

jayrajput:
I have made changes to the setup.py to include the auto completion script in the python package. I am not sure if I need to do that. Also if the version for the mozdownloand in the setup.py needs to be updated.


Nebelhom:
@whimboo , @jayrajput 

Hi folks! Spring cleaning ;) What is the status of this PR? Can it be closed or will we work on it in the future?

jayrajput:
@Nebelhom, @whimboo 

The idea of using tab complete is not feasible with the optcomplete as the optcomplete module calls the mozdownload to find the completion and mozdownload takes 2-3 seconds to import all the needed modules. The slowest module to load is requests. Please see my previous comments for details.

Loading the modules conditionally is one possibility, but that will start getting tricky.
Any suggestion on how to implement this? I can start working on it.

whimboo:
Sorry for totally forgetting about this PR! And thanks to @Nebelhom to bring it back to my attention. I can now see the problems we would have here. That's really unfortunate. I wonder if we shouldn't import the requests module globally but only when we really need it. That would drastically help. I would suggest we investigate that in a new issue first. It can block issue #22.

Given that you have removed your fork you won't be able to push to this PR anymore. So lets close it.

Issue #215, opened 2014-05-26
`mac` platform option for tinderbox builds should default to `macosx64`


whimboo:
I think we can get rid of `mac64` as option. I don't see any tinderbox builds existent for 32bit only, even not in the long-term archive: http://inbound-archive.pub.build.mozilla.org/pub/mozilla.org/firefox/tinderbox-builds/.

So `mac64` should be removed, and `mac` should become the default.

whimboo:
@Nebelhom it looks like the summary was not that good! Sorry, that it makes it hard to understand for you. I updated the summary now, and hope it's better.

whimboo:
PR #235 has been merged. Thanks!

PR #235, opened 2015-01-16
Make mac and mac64 platform equivalent for tinderbox builds (#215)

Nebelhom:
This PR addresses issue #215 

From what I can gather, this issue is a simple deletion job, so I just took out the necessary parts in `scraper.py` and in the tests.

### DIFF
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -756,7 +756,6 @@ class TinderboxScraper(Scraper):
         regex_suffix = {'linux': r'.*\.%(EXT)s$',
                         'linux64': r'.*\.%(EXT)s$',
                         'mac': r'.*\.%(EXT)s$',
-                        'mac64': r'.*\.%(EXT)s$',
                         'win32': r'.*(\.installer%(STUB)s)\.%(EXT)s$',
                         'win64': r'.*(\.installer%(STUB)s)\.%(EXT)s$'}
 
@@ -863,7 +862,6 @@ class TinderboxScraper(Scraper):
         PLATFORM_FRAGMENTS = {'linux': 'linux',
                               'linux64': 'linux64',
                               'mac': 'macosx',
-                              'mac64': 'macosx64',
                               'win32': 'win32',
                               'win64': 'win64'}
 
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -863,7 +863,7 @@ class TinderboxScraper(Scraper):
         PLATFORM_FRAGMENTS = {'linux': 'linux',
                               'linux64': 'linux64',
                               'mac': r'macosx(64)?',
-                              'mac64': r'macosx(64)?',
+                              'mac64': 'macosx64',
                               'win32': 'win32',
                               'win64': 'win64'}
 
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -862,7 +862,7 @@ class TinderboxScraper(Scraper):
 
         PLATFORM_FRAGMENTS = {'linux': 'linux',
                               'linux64': 'linux64',
-                              'mac': r'macosx(64)?',
+                              'mac': 'macosx64',
                               'mac64': 'macosx64',
                               'win32': 'win32',
                               'win64': 'win64'}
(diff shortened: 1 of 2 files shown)


Nebelhom:
@whimboo @davehunt 

woops. Sorry, forgot to mention. Please check it...

mozdownload/scraper.py
```diff
@@ -863,7 +862,6 @@ def platform_regex(self):
         PLATFORM_FRAGMENTS = {'linux': 'linux',
                               'linux64': 'linux64',
                               'mac': 'macosx',
-                              'mac64': 'macosx64',
```
whimboo: Not sure why you are getting this portion removed. We only have macosx64 folders for tinderbox builds like `mozilla-central-macosx64.*/`. It would be better to say that `mac` uses those too, if not done yet.

Nebelhom: @whimboo 

Maybe I misread your statement from the issue (link #215). I read it as: "Remove mac64 option"... so I did ;)

I did not give the option in the regex of to download either macosx or macosx64 because I thought this might lead to a bug or inaccuracies (not possible to download a specific build because of it...)

whimboo: There is only one set of builds we have for Firefox. It's the bundle with 32bit/64bit binaries. The old bundle with PowerPC/32bit is not shipped anymore. So `mac` and `mac64` should point to the same bundle.

I may have written it the wrong way in issue #215. For consistency we should just keep mac64, just in case that we do not break any downstream projects.

Nebelhom:
@whimboo I hope that is now what you meant, but I fear I misunderstood you yet again ;)

Thanks for looking at it.

mozdownload/scraper.py
```diff
@@ -862,8 +862,8 @@ def platform_regex(self):
 
         PLATFORM_FRAGMENTS = {'linux': 'linux',
                               'linux64': 'linux64',
-                              'mac': 'macosx',
-                              'mac64': 'macosx64',
+                              'mac': r'macosx(64)?',
+                              'mac64': r'macosx(64)?',
```
whimboo: We don't need optional brackets here. For tinderbox builds there are only `macosx64` subfolders.

tests/tinderbox_scraper/test_tinderbox_scraper.py
```diff
@@ -63,13 +63,6 @@
      'target': 'mozilla-central-firefox-25.0a1.en-US.win64-x86_64.installer.exe',
      'target_url': 'firefox/tinderbox-builds/mozilla-central-win64/'
                    '1374583608/firefox-25.0a1.en-US.win64-x86_64.installer.exe'},
-    # -a firefox -p mac64 --branch=mozilla-central
-    {'args': {'application': 'firefox',
-              'branch': 'mozilla-central',
-              'platform': 'mac64'},
-     'target': 'mozilla-central-firefox-25.0a1.en-US.mac.dmg',
-     'target_url': 'firefox/tinderbox-builds/mozilla-central-macosx64/'
-                   '1374583608/firefox-25.0a1.en-US.mac.dmg'},
```
whimboo: We leave mac64 as an option, so it would be good to keep those tests.

Nebelhom:
@whimboo: Let's hope we got it right, this time ;) please review

mozdownload/scraper.py
```diff
@@ -862,7 +862,7 @@ def platform_regex(self):
 
         PLATFORM_FRAGMENTS = {'linux': 'linux',
                               'linux64': 'linux64',
-                              'mac': 'macosx',
+                              'mac': r'macosx(64)?',
```
whimboo: As mentioned in my last review, the brackets are not necessary. That does not only apply to `mac64` but also `mac`.

Nebelhom:
@whimboo Let's try this one. Sorry for making this simple fix so long-winded. I entirely misunderstood this issue...

whimboo:
No worries @Nebelhom! My original summary was even not that clear and probably caused the confusion. This PR finally landed now as https://github.com/mozilla/mozdownload/commit/2fec0598ac8816ded72110f82120bd36245381ea

Issue #236, opened 2015-01-18
invalid branch tests in DailyScraper Tests


Nebelhom:
I noticed that there are still some error messages to be tested in the manual `test_cases.md`.

This issue addresses the those testing the error message involved when passing an invalid branch to `DailyScraper`

whimboo:
PR #237 has been merged.

PR #237, opened 2015-01-18
Added tests checking the correct error message when passing an invalid branch (#236)

Nebelhom:
This PR addresses issue #236

### DIFF
--- a/tests/daily_scraper/test_invalid_branch.py
+++ b/tests/daily_scraper/test_invalid_branch.py
@@ -21,19 +21,6 @@ tests_with_invalid_branch = [
     {'args': {'application': 'firefox',
               'branch': 'invalid',
               'platform': 'win32'}
-     },
-    # -a firefox -t daily -p win32 --branch=invalid --retry-attempts=2
-    {'args': {'application': 'firefox',
-              'branch': 'invalid',
-              'platform': 'win32',
-              'retry_attempts': 2}
-     },
-    # -a firefox -t daily -p win32 --branch=invalid --retry-attempts=2 --retry-delay=0
-    {'args': {'application': 'firefox',
-              'branch': 'invalid',
-              'platform': 'win32',
-              'retry_attempts': 2,
-              'retry_delay': 0.}
      }
 ]
 
--- a/tests/daily_scraper/test_invalid_branch.py
+++ b/tests/daily_scraper/test_invalid_branch.py
@@ -4,11 +4,8 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
-import os
 import mozlog
-import sys
 import unittest
-import urllib
 
 from mozdownload import DailyScraper, NotFoundError
 from mozdownload.utils import urljoin
@@ -27,7 +24,7 @@ tests_with_invalid_branch = [
 tests = tests_with_invalid_branch
 
 
-class TestDailyScraper_invalidBranchParameter(mhttpd.MozHttpdBaseTest):
+class TestDailyScraper(mhttpd.MozHttpdBaseTest):
     """test mozdownload DailyScraper class with invalid branch parameters"""
 
     def test_scraper(self):
--- a/tests/daily_scraper/test_invalid_branch.py
+++ b/tests/daily_scraper/test_invalid_branch.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import os
+import mozlog
+import sys
+import unittest
+import urllib
+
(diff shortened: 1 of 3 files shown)


tests/daily_scraper/test_invalid_branch.py
```diff
@@ -0,0 +1,59 @@
...
+              'branch': 'invalid',
+              'platform': 'win32',
+              'retry_attempts': 2}
+     },
+    # -a firefox -t daily -p win32 --branch=invalid --retry-attempts=2 --retry-delay=0
+    {'args': {'application': 'firefox',
+              'branch': 'invalid',
+              'platform': 'win32',
+              'retry_attempts': 2,
+              'retry_delay': 0.}
```
whimboo: I wonder why we have to use the retry logic for invalid branches. It won't help anyway. I think retry we should do in its very own test module.

Nebelhom: When I ran the tests manually to see the result, I had the same thoughts, but went along with it anyways, because we had it in the list.

If you could let me know what the significance of the test with `retry_delay = 0.`, I would be grateful :)

whimboo: The list was not that accurate then. ;) Retries will make sense when the content to be downloaded will become available later. I would suggest we get a separate test module up for that, and maybe we can tweak mozhttpd so it serves a given content after x number of tries.

whimboo: > If you could let me know what the significance of the test with `retry_delay = 0.`, I would be grateful :)

It means retry immediately and do not wait x seconds (i think default 10s)

Nebelhom:
@whimboo 

And here as well. Please check the new commit. Thank you :)

tests/daily_scraper/test_invalid_branch.py
```diff
@@ -0,0 +1,46 @@
...
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import os
+import mozlog
+import sys
+import unittest
+import urllib
```
whimboo: Some of those imported modules are not used. Please get rid of them.

tests/daily_scraper/test_invalid_branch.py
```diff
@@ -0,0 +1,46 @@
...
+    {'args': {'application': 'firefox',
+              'branch': 'invalid',
+              'platform': 'win32'}
+     }
+]
+
+tests = tests_with_invalid_branch
+
+
+class TestDailyScraper_invalidBranchParameter(mhttpd.MozHttpdBaseTest):
```
whimboo: The class name is broken here. It should be `TestDailyScraper`. The test method can have details about what we actually test. We should also fix all the other test files.

tests/test_cases.md
```diff
@@ -36,11 +36,14 @@ mozdownload --url=https://mozqa.com/index.html -d url-downloads
 mozdownload --url=https://mozqa.com/index.html --username=username --password=password
```
whimboo: Not directly related but we can remove all of those tests. Those have been automated now.

tests/daily_scraper/test_invalid_branch.py
Nebelhom: Do you want to have the name changes in the other files done in this patch, as well, or in a separate one?

I would opt for a separate one, just in case people try to retrace the steps

tests/test_cases.md
Nebelhom: Sorry to pretend to be silly, but which tests exactly?

Retries, I just put in, so you can't mean those. The password username one? All the error ones?

Again sorry, just making sure that I don't have to revert out of a patch because I misunderstood

tests/daily_scraper/test_invalid_branch.py
whimboo: Yeah, move it out to a separate issue. We can make it mentored even.

tests/test_cases.md
whimboo: It's about the DirectScraper with BasicAuth. We landed tests for all that recently.

Nebelhom:
@whimboo 

I removed the tests now. Hopefully that should be it from that. Please review. Thanks :)

tests/daily_scraper/test_invalid_branch.py
```diff
@@ -0,0 +1,43 @@
...
+     }
+]
+
+tests = tests_with_invalid_branch
+
+
+class TestDailyScraper(mhttpd.MozHttpdBaseTest):
+    """test mozdownload DailyScraper class with invalid branch parameters"""
+
+    def test_scraper(self):
```
whimboo: The method should be named like `test_invalid_branch`. With that fixed it looks fine and I will get it merged.

Nebelhom:
@whimboo 

I addressed your change suggestions. Should be ready to get merged now. Please review :)

whimboo:
Thanks Johannes! I merged your PR as https://github.com/mozilla/mozdownload/commit/d2abf381bd1f3c08b1037c3665005bb779eb1b09.

Issue #143, opened 2013-08-27
TinderboxScraper: Omit folders without binaries


Nebelhom:
Issue #11 implemented in the DailyScraper that empty folders are omitted from the binary search.

We would need this for the TinderboxScraper, too.

Here is a current test and its traceback:

`mozdownload -a firefox -t tinderbox -p linux --branch=mozilla-central`

`Retrieving list of builds from https://ftp.mozilla.org/pub/mozilla.org/firefox/tinderbox-builds/mozilla-central-linux
Found 209 builds: 1374305748, 1374398028, 1374400923, 1374487346, 1374500811 ... 1377562603, 1377562784, 1377569268, 1377583063, 1377597721
Traceback (most recent call last):
  File "scraper.py", line 960, in <module>
    cli()
  File "scraper.py", line 957, in cli
    build.download()
  File "scraper.py", line 237, in download
    if os.path.isfile(os.path.abspath(self.target)):
  File "scraper.py", line 197, in target
    self.build_filename(self.binary))
  File "scraper.py", line 148, in binary
    self.path)
__main__.NotFoundError: Binary not found in folder: https://ftp.mozilla.org/pub/mozilla.org/firefox/tinderbox-builds/mozilla-central-linux/1377597721`

A related issue is also the current failure to download a file with the `txt` extension which is also most likely down to this problem.

Nebelhom:
@whimboo 

I am looking for a good test case, but I cannot find a recent folder that is still empty.

Any suggestions?

whimboo:
This might be very hard to find. Maybe directly create testcases for mozhttpd?

Nebelhom:
@whimboo I was worried about that option. It means that I have to rely on a test case that I wrote ;)

It's ok though, I managed to mimic it now. Let's see if I get anywhere with it.

whimboo:
PR #242 has been merged and this will be in the 1.13 release now!

Issue #11, opened 2012-05-23
If folder of latest version doesn't contain binary go back in history for that day


davehunt:
When attempting to download Thunderbird daily builds the following exception is thrown:

```
$ mozdownload --application=thunderbird --type=daily --branch=comm-central --date=2012-05-23
Retrieving list of builds from https://ftp.mozilla.org/pub/mozilla.org/thunderbird/nightly/2012/05/
Traceback (most recent call last):
  File "/Users/dhunt/.virtualenvs/8690aa596f38bf36/bin/mozdownload", line 8, in <module>
    load_entry_point('mozdownload==1.0', 'console_scripts', 'mozdownload')()
  File "/Users/dhunt/.virtualenvs/8690aa596f38bf36/lib/python2.7/site-packages/mozdownload-1.0-py2.7.egg/mozdownload/scraper.py", line 744, in cli
    build.download()
  File "/Users/dhunt/.virtualenvs/8690aa596f38bf36/lib/python2.7/site-packages/mozdownload-1.0-py2.7.egg/mozdownload/scraper.py", line 173, in download
    if os.path.isfile(os.path.abspath(self.target)):
  File "/Users/dhunt/.virtualenvs/8690aa596f38bf36/lib/python2.7/site-packages/mozdownload-1.0-py2.7.egg/mozdownload/scraper.py", line 143, in target
    self.build_filename(self.binary))
  File "/Users/dhunt/.virtualenvs/8690aa596f38bf36/lib/python2.7/site-packages/mozdownload-1.0-py2.7.egg/mozdownload/scraper.py", line 84, in binary
    raise NotFoundException("Binary not found in folder", self.path)
mozdownload.scraper.NotFoundException: Binary not found in folder: https://ftp.mozilla.org/pub/mozilla.org/thunderbird/nightly/2012/05/2012-05-23-03-38-23-comm-central
```

It appears that the folder that's selected only contains test results and not the builds.

PR #242, opened 2015-01-31
Tinderbox builds: If latest folder is empty go to newest folder with build (#143)

Nebelhom:
This PR addresses issue #143 

I essentially followed the fix for Issue #11 and transferred the `DailyScraper`'s `self.is_buid_dir` function to the `TinderboxScraper`.

I left the line starting with `self.application...` in, in case we will at one point fully support b2g downloads.

### DIFF
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -425,6 +425,8 @@ class DailyScraper(Scraper):
     def is_build_dir(self, dir):
         """Return whether or not the given dir contains a build."""
 
+        # Cannot move up to base scraper due to parser.entries call in
+        # get_build_info_for_date (see below)
         url = urljoin(self.base_url, self.monthly_build_list_regex, dir)
 
         if self.application in MULTI_LOCALE_APPLICATIONS \
@@ -816,6 +818,30 @@ class TinderboxScraper(Scraper):
 
         return platform
 
+    def is_build_dir(self, dir):
+        """Return whether or not the given dir contains a build."""
+
+        # Cannot move up to base scraper due to parser.entries call in
+        # get_build_info_for_index (see below)
+        url = urljoin(self.base_url, self.build_list_regex, dir)
+
+        if self.application in MULTI_LOCALE_APPLICATIONS \
+                and self.locale != 'multi':
+            url = urljoin(url, self.locale)
+
+        parser = DirectoryParser(url, authentication=self.authentication,
+                                 timeout=self.timeout_network)
+
+        pattern = re.compile(self.binary_regex, re.IGNORECASE)
+        for entry in parser.entries:
+            try:
+                pattern.match(entry).group()
+                return True
+            except:
+                # No match, continue with next entry
+                continue
+        return False
+
     def get_build_info_for_index(self, build_index=None):
         url = urljoin(self.base_url, self.build_list_regex)
 
@@ -823,6 +849,7 @@ class TinderboxScraper(Scraper):
         parser = DirectoryParser(url, authentication=self.authentication,
                                  timeout=self.timeout_network)
         parser.entries = parser.filter(r'^\d+$')
+        parser.entries = parser.filter(self.is_build_dir)
 
         if self.timestamp:
             # If a timestamp is given, retrieve the folder with the timestamp
--- a/tests/data/firefox/tinderbox-builds/mozilla-central-linux/1374583609/README.md
+++ b/tests/data/firefox/tinderbox-builds/mozilla-central-linux/1374583609/README.md
@@ -1,0 +1,3 @@
+This is supposed to be an empty directory to serve as a test case for a missing binary.
+
+See Issue #143 "TinderboxScraper: Omit folders without binaries"


Nebelhom:
@whimboo @davehunt 

Could one of you please review the pull? Thank you :)

mozdownload/scraper.py
```diff
@@ -816,6 +816,28 @@ def detect_platform(self):
...
         return platform
 
+    def is_build_dir(self, dir):
+        """Return whether or not the given dir contains a build."""
+
+        url = urljoin(self.base_url, self.build_list_regex, dir)
+
+        if self.application in MULTI_LOCALE_APPLICATIONS \
+                and self.locale != 'multi':
+            url = urljoin(url, self.locale)
```
whimboo: Why do we have to add the locale again to the URL? Isn't it already part of `build_list_regex`?

mozdownload/scraper.py
```diff
@@ -816,6 +816,28 @@ def detect_platform(self):
...
+
+        pattern = re.compile(self.binary_regex, re.IGNORECASE)
+        for entry in parser.entries:
+            try:
+                pattern.match(entry).group()
+                return True
+            except:
+                # No match, continue with next entry
+                continue
+        return False
```
whimboo: When looking at this method and the existing is_build_dir method at https://github.com/mozilla/mozdownload/blob/master/mozdownload/scraper.py#L425, I would suggest you move all the code up to the base scraper class, and let it require the URL as parameter. That way we can make sure that all type of build has that method. It doesn't look different to me code-wise.

mozdownload/scraper.py
```diff
@@ -823,6 +845,7 @@ def get_build_info_for_index(self, build_index=None):
         parser = DirectoryParser(url, authentication=self.authentication,
                                  timeout=self.timeout_network)
         parser.entries = parser.filter(r'^\d+$')
+        parser.entries = parser.filter(self.is_build_dir)
```
whimboo: Hm, `is_build_dir` should return a boolean when I look at its name. Maybe we can rename it right away to something better? Especially when we want to also download other types as builds, the name would not match.

Feel free to file a new issue for that.

mozdownload/scraper.py
Nebelhom: Isn't it because the path for applications with multi is slightly different and cannot be put into one function or some such thing? I don't know tbh it was a copy and paste...

Nebelhom: hmmm... trying to download a tinderbox build of b2g gave me the following error both in the PR (with commenting out of the section and without) and on the current master branch

>> /mozdownload$ ./scraper.py -a b2g -t tinderbox
  File "./scraper.py", line 1075, in <module>
    cli()
  File "./scraper.py", line 1070, in cli
    build.download()
  File "./scraper.py", line 281, in download
    if os.path.isfile(os.path.abspath(self.target)):
  File "./scraper.py", line 241, in target
    self.build_filename(self.binary))
  File "./scraper.py", line 199, in binary
    raise NotFoundError(message, self.path)
__main__.NotFoundError: Specified build has not been found: https://ftp.mozilla.org/pub/mozilla.org/b2g/tinderbox-builds/mozilla-central-l10n

So unless I have made a mistake in usage of the scraper for b2g (a MULTI_APPLICATION), this part of the code is broken and needs fixing... but that would be beyond the scope of this PR.

What is your take on it @whimboo. Taking the mentioned section out or leaving it in?

mozdownload/scraper.py
Nebelhom: will do.

mozdownload/scraper.py
Nebelhom: At one point I thought the same... I don't know why I didn't do it... brain failure, maybe...

Thanks for pointing it out.

mozdownload/scraper.py
whimboo: I don't think we have tinderbox support for b2g yet. So simply ignore this specific case.

whimboo:
Hey Johannes, will you be able to get an update by today? We want to release 1.13 soonish and this fix looks like a great enhancement, which we could also ship.

mozdownload/scraper.py
Nebelhom: @whimboo 
I now remember why I could not do it.

Please see https://github.com/mozilla/mozdownload/blob/master/mozdownload/scraper.py#L465 

This call to `parser.filter` prevents us from moving it up to base scraper. Maybe there is a way to circumvent this, but it may not look very clean.

As I mentioned in IRC, I once or twice used the built-in functools.partial method to cirumvent the constraints of only one argument. Don't know if it can work here, too.

mozdownload/scraper.py
Nebelhom: To sum it up for all. We left this part in, as it only affects b2g builds at the moment, which we do not officially support. In case removing this part leads to a regression, we leave it for now.

mozdownload/scraper.py
Nebelhom: See #245

Nebelhom:
@whimboo 

I ma the changes agreed upon, so nothing except the comments ;) and squashed the commits into one. Should be good to go now

mozdownload/scraper.py
whimboo: As agreed lets skip it for now and leave all as it is.

mozdownload/scraper.py
whimboo: So specific locales here only en-US is available in a sub folder. The binary including all other locales is in the same folder. That's why the locale part has to be added again. Keep in mind it works only for en-US at the moment. :)

tests/data/firefox/tinderbox-builds/mozilla-central-linux/1374583609/README.md
```diff
@@ -0,0 +1,3 @@
+This is supposed to be an empty directory to serve as a test case for a missing binary.
+
+See Issue #143 "TinderboxScraper: Omit folders without binaries"
```
whimboo: Shouldn't we add a specific test for it?

Nebelhom: @whimboo 

Well this is a specific test for it ;)

For this test: https://github.com/mozilla/mozdownload/blob/master/tests/tinderbox_scraper/test_tinderbox_scraper.py#L46

the empty folder would be chosen, since it is the latest build. But the PR allows it now to check if there is actually something inside the directory first.

I just extended the original test, by adding to the test directory structure.

mozdownload/scraper.py
Nebelhom: @whimboo I will keep it mind, but does this mean I have to change anything in the current PR? I wouldn't...

tests/data/firefox/tinderbox-builds/mozilla-central-linux/1374583609/README.md
whimboo: But that is using latest build. So whenever someone of us will add more entries to this folder, this specific test will be invalid. Or does this behavior only work for the request to get the latest build?

whimboo: Hm, that may it. I was thinking too complex. Ok then.

mozdownload/scraper.py
whimboo: Nope, nothing for this bug. Don't be afraid! :)

whimboo:
I merged this PR as https://github.com/mozilla/mozdownload/commit/c73e81463c11cd9667d58322390fa99b6a35b135. Thanks Johannes!

tests/data/firefox/tinderbox-builds/mozilla-central-linux/1374583609/README.md
Nebelhom: I can only do simple. You do the complex thinking ;)

Issue #214, opened 2014-05-26
test_direct_scraper.py should make use of local http server and not mozqa.com


whimboo:
Right now the given tests wants to download a file from mozqa.com. It should really get if from the locally served pages via mozhttpd.

jasonwist:
Yeah, it's been a while since I was looking at this (new job + being sick). 

Where can I find mozhttpd? I can see it being imported/used in mozhttpd_base_test but I can't find the where it (mozhttpd) and it's methods are defined.

whimboo:
The document root of the server always points to `/tests/data`, so what you need in the test is self.wdir as defined here: https://github.com/mozilla/mozdownload/blob/master/tests/mozhttpd_base_test.py#L43

Here an example in how to use the local web server for serving content:
https://github.com/mozilla/mozdownload/blob/master/tests/base_scraper/test_base_scraper.py#L33

Please let me know if you have further questions.

jasonwist:
How can I test to see if my changes have worked?

whimboo:
There is a `run_tests.py` script in the root folder of the repository. Just run it and you will see if all the tests will pass, or if a failure is around for your changes.

jasonwist:
I tried to inherit self.wdir from mhttpd_base_test (how I assume BaseScraperTest was getting access to it) but that wasn't working, so now I'm implementing all the steps to get wdir in TestDirectScraper. 

Was I on the right path before?

Also why does urljoin() use 'download_test.txt' in BaseScraperTest? Is that the name of a file we are downloading?

whimboo:
`self.wdir` will automatically be available if you inherit the test case class from mhttpd.MozHttpdBaseTest. So you can just use it as in BaseScraperTest:

https://github.com/mozilla/mozdownload/blob/master/tests/base_scraper/test_base_scraper.py#L35

urljoin() is indeed used to specify the url to a locally hosted test file. You can find it here: https://github.com/mozilla/mozdownload/blob/master/tests/data/download_test.txt

So the code for the direct scraper test can mostly look totally the same in terms of navigating to the URL as for the base scraper test.

jasonwist:
I still want to finish it. Just prepping for a big job interview next week.

whimboo:
PR #312 has been merged to master.

PR #244, opened 2015-02-05
Issue #214 mozhttpd download

jasonwist:
Hi,

I tried the git commands you suggested. I think they worked.

Here is what I got to before I had posted a few days ago.

Is this the best way to show you what I have been working on/what I am having troubles with?


tests/direct_scraper/test_direct_scraper.py
```diff
@@ -23,7 +26,7 @@ def tearDown(self):
         mozfile.rmtree(self.temp_dir)
 
     def test_url_download(self):
-        test_url = 'https://mozqa.com/index.html'
+        test_url = urljoin(self.wdir, 'download_test.txt')
```
whimboo: The problem why this is not working is that `setUp()` is not calling itself via the base class. But given that the code for creating the tmp folder is already present in `MozHttpdBaseTest.setUp` you can get rid of the whole method. The same applies also for `tearDown`.

.gitignore
```diff
@@ -2,4 +2,4 @@
 *.pyc
 build
 dist
-tests/venv
+tests/venv
```
whimboo: nit: this change can be reverted.

whimboo:
Btw. this will become the fix for issue #214.

Nebelhom:
Hi @jasonwist ,

are you still working on this PR?

If you are stuck with it, just write a comment an we will see if we can help you out :)

whimboo:
Going to close this commit given that we haven't seen a response in the last 3 months. Feel free to reopen if you think that you will continue. Thanks.

Issue #243, opened 2015-02-05
Cannot download win64 bit builds of Firefox anymore


andreeamatei:
Our testruns failed today on win64 nodes, not being able to find today's build.
This command will show the error:
mozdownload --type=daily --branch=mozilla-central --platform=win64 --locale=en-US --build-id=20150205030205
While 32bit works.

07:13:54 [mozilla-central_functional] $ cmd.exe /C '"mozmill-env-%ENV_PLATFORM%\run mozdownload --type=%BUILD_TYPE% --branch=mozilla-central --platform=%PLATFORM% --locale=%LOCALE% --build-id=%BUILD_ID% --date=%DATE% --retry-attempts=10 --retry-delay=30 --directory=builds && exit %%ERRORLEVEL%%"'
07:13:55   INFO | Retrieving list of builds from https://ftp.mozilla.org/pub/mozilla.org/firefox/nightly/2015/02
07:13:55   WARNING | Build not found: 'Folder for builds on 2015-02-05-03-02-05 has not been found: https://ftp.mozilla.org/pub/mozilla.org/firefox/nightly/2015/02'
07:13:55   INFO    | Will retry in 30.0 seconds...

Related to bug https://bugzilla.mozilla.org/show_bug.cgi?id=1128953

whimboo:
I will work on that. Updating the tests take a bit more time here, so a PR should be around tomorrow. I will temporarily disable testing win64 builds in mozmill-ci production.

whimboo:
PR #248 got merged.

PR #246, opened 2015-02-09
Update scraper to pull win64 files instead of win64-x86_64. Fixes issue ...

AutomatedTester:
...#243


mozdownload/scraper.py
```diff
@@ -46,7 +46,7 @@
                       'mac': r'mac',
                       'mac64': r'mac(64)?',
                       'win32': r'win32',
-                      'win64': r'win64-x86_64'}
```
whimboo: Thank you for the patch David, but as I have mentioned on IRC yesterday I have a nearly ready patch to go too. Something else which makes it a bit more complicate is that we cannot hard-switch as you did here. We have to support both naming schemes because Thunderbird didn't flip that yet, and we would not be able to download older Firefox 64 bit builds. I hope you have nothing against it when I get my PR finished instead?

whimboo:
PR #248 is the complete one. I'm closing this PR in favor of the other one.

Issue #249, opened 2015-02-10
Release version 1.13


whimboo:
We want to release version 1.13 of mozdownload.

The following tasks have to be performed:
- [x] Update setup.py for version bump
- [x] Update changelog
- [ ] Tag version 1.13
- [ ] Release to pypi

whimboo:
Tagged as 1.13:
https://github.com/mozilla/mozdownload/commits/1.13

Released to pypi:
Submitting dist/mozdownload-1.13.tar.gz to https://pypi.python.org/pypi
Server response (200): OK

PR #251, opened 2015-02-11
Release version 1.13 (#249)

whimboo:
This is the PR necessary for issue #249. Just a version bump and changes to the history file.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@
 
 
 import os
-from setuptools import setup, find_packages
+from setuptools import setup
 
 try:
     here = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -6,7 +6,7 @@
 
 
 import os
-from setuptools import setup, find_packages
```
Nebelhom: Why are we removing find_packages?

Nebelhom:
@whimboo 

I cannot see much wrong with it. Just answer the remark satisfactorily and you will have a +1 from me.

setup.py
whimboo: It was never used in this file, so just an orphaned include.

whimboo:
This PR has been merged as https://github.com/mozilla/mozdownload/commit/ea791a65f2900a3923e01b46b8cc88fabd6d5838

Issue #253, opened 2015-02-12
Large regression in tinderbox download times


EricRahm:
It looks like c73e81463c11cd9667d58322390fa99b6a35b135 caused a very large regression in the amount of time it takes to lookup a tinderbox build. The `is_build_dir` check needs to be moved, the simple solution would be to move it to *after* we filter by timestamp / date.

whimboo:
Thanks Eric! So it is a regression from issue #143. If that regression is important for you to get fixed we could do it quickly and get a follow-up release out.

EricRahm:
@whimboo yes, it's causing several minute delays per build.

whimboo:
Definitely something we want to have for the 1.14 release.

whimboo:
PR #254 has been merged! Thank you @EricRahm for fixing it!

whimboo:
It looks like we may have a problem with index==0. We have to check that.

whimboo:
I checked this myself and it's only an issue of what we actually print out as failure message for the URL. It's not critical and if necessary can be fixed in another issue. For details see: https://github.com/mozilla/mozdownload/pull/254/files#r25765397. Closing out this issue.

PR #254, opened 2015-02-12
Limit checks for non-empty build dirs (#253)

EricRahm:
Checking if a build dir is empty is an expensive operation. This
iteratively checks for a non-empty build dir only when necessary and
stops checking when the first non-empty build dir is found.

### DIFF
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -872,7 +872,6 @@ class TinderboxScraper(Scraper):
         parser = DirectoryParser(url, authentication=self.authentication,
                                  timeout=self.timeout_network)
         parser.entries = parser.filter(r'^\d+$')
-        parser.entries = parser.filter(self.is_build_dir)
 
         if self.timestamp:
             # If a timestamp is given, retrieve the folder with the timestamp
@@ -892,7 +891,13 @@ class TinderboxScraper(Scraper):
 
         # If no index has been given, set it to the last build of the day.
         if build_index is None:
-            build_index = len(parser.entries) - 1
+            # Find the most recent non-empty entry.
+            build_index = len(parser.entries)
+            for build in reversed(parser.entries):
+                build_index -= 1
+                if not build_index or self.is_build_dir(build):
+                    break
+
         self.logger.info('Selected build: %s' % parser.entries[build_index])
 
         return (parser.entries, build_index)


whimboo:
Haven't seen this PR yet given that I wasn't mentioned or the issue has been directly referenced. So this is a fix for issue #253.

whimboo:
@Nebelhom, I'm out for a couple days and not back before Wednesday. Would you mind to have a look at Eric's PR? Maybe you can sort out any questions, if some are coming up. We can see about a follow-up release next week.

EricRahm:
Just for perspective, there are over 1000 entries in the linux64 m-i directory. Try running:
    `mozdownload -t tinderbox -p linux64 --branch mozilla-inbound`

mozdownload/scraper.py
```diff
@@ -892,7 +891,13 @@ def get_build_info_for_index(self, build_index=None):
 
         # If no index has been given, set it to the last build of the day.
         if build_index is None:
-            build_index = len(parser.entries) - 1
+            # Find the most recent non-empty entry.
+            build_index = len(parser.entries)
+            for build in reversed(parser.entries):
```
Nebelhom: Could you maybe explain, why you are doing it this particular way? I only program as a hobby, so this may be an approach that is very common, but I have not seen it before.

In your `for` loop, you create a list that is a reversed `parser.entries`, and then you take `build_index` and count down from there.

Why don't you use `enumerate` and get the index this way and then only check for `is_build_dir`?

Sorry for asking such a silly question, but it took me a bit to figure out what you were doing and I was wondering why. Maybe I am also missing something and I can learn something here...

Nebelhom:
@whimboo I'll try my best.

@EricRahm I have noted down one question, which is for my understanding, so I would be very glad if you could enlighten me (Sorry for being a pain ;) )

As a side note, I suggest we do the same change in a different issue for the DailyScraper `is_build_dir` (since #253 is only defined for tinderbox)

Even if it is maybe only a small issue for nightly, I find it good coding practice to optimise our code when we are aware of an issue.

mozdownload/scraper.py
EricRahm: No worries, these are good questions and mostly related to personal style.

As to why I'm using `reversed`: we only care about the entries at the end (the latest entries), so we want to start with those first. So rather than processing over 1000 entries, hopefully we just process 1.

As to why not enumerate: `reversed` takes a sequence and returns an iterator and `enumerate` actually returns an enumerator (not a sequence). So you'd have to convert the enumerator back to a list. Something like:
    `for build_index, build in reversed(list(enumerate(parser.entries))):` 
Which seems both unintuitive and wasteful for a list of > 1000 entries. Of course I could be convinced otherwise!

Thanks for the review.

Nebelhom: Thanks for the explanation. It makes perfect sense now. I find it helpful to see how other people operate. Gives me a different outlook on problems to solve in the future.

Cheers :)

mozdownload/scraper.py
```diff
@@ -892,7 +891,13 @@ def get_build_info_for_index(self, build_index=None):
...
+                build_index -= 1
+                if not build_index or self.is_build_dir(build):
+                    break
+
```
whimboo: This looks fine to me but I miss a NotFoundError thrown as in line 888 in case of no build has been found. Right now we would return an invalid build index. I think we can add a test for that simply via `mozdownload -t tinderbox -p linux64 --branch mozilla-inbound --ext rr` because there won't exist such a file.

whimboo:
@EricRahm do you have a chance to get this PR updated? I would like to get this and some other PR merged soon for a new release. Thanks.

EricRahm:
@whimboo We'd still get a NotFoundError when [trying to download](https://github.com/mozilla/mozdownload/blob/master/mozdownload/scraper.py#L159), does that suffice?

whimboo:
I was testing it again and I'm getting the following error: `requests.exceptions.SSLError: EOF occurred in violation of protocol (_ssl.c:581)` after about 2 minutes. Maybe I exceeded the request limit or such. But this also appears when I run all without your patch. 

So I think we are fine here and can get this landed.

whimboo:
I merged the PR as https://github.com/mozilla/mozdownload/commit/741836dc298d086e6df5fff5b1c107f859535239.

mozdownload/scraper.py
sydvicious: On line 898, whiy would build_index ever be false or None?

sydvicious:
I don't see this patch in master.

mozdownload/scraper.py
EricRahm: @sydvicious build_index can be zero which is a falsy value.

sydvicious: But 0 is a valid index anyway. You don't want to assume that the directory at index 0 is OK. If the directory at index 0 is not a good build directory, this will return it anyway.

whimboo: If there is a problem with it please file a new PR and at least re-open the issue but not the pull request.

whimboo: > But 0 is a valid index anyway. You don't want to assume that the directory at index 0 is OK. If the directory at index 0 is not a good build directory, this will return it anyway.

I just checked this myself given that I haven't gotten a response from @EricRahm yet on this PR or issue. So we indeed return the oldest build folder for tinderbox builds but a download will fail with a NotFoundError anyway. It's only a matter of the message we print. That's all. We could improve that later by printing the platform build folder instead of the first build folder under it.

Issue #258, opened 2015-02-23
 mozdownload cannot load tinderbox zip builds for Windows


sydvicious:
mozdownload --type=tinderbox --platform=win32 --extension=zip --branch=mozilla-beta

The regular expression at line 760 ends up looking for firefox*.en-US.win32.installer.zip, which it won't find.

        regex_base_name = r'^%(APP)s-.*\.%(LOCALE)s\.%(PLATFORM)s'
        regex_suffix = {'linux': r'.*\.%(EXT)s$',
                        'linux64': r'.*\.%(EXT)s$',
                        'mac': r'.*\.%(EXT)s$',
                        'mac64': r'.*\.%(EXT)s$',
                        'win32': r'.*(\.installer%(STUB)s)\.%(EXT)s$',
                        'win64': r'.*(\.installer%(STUB)s)\.%(EXT)s$'}

whimboo:
We want this fix in the next release 1.14.

sydvicious:
Supercded by #264

PR #259, opened 2015-02-23
 mozdownload cannot load tinderbox zip builds for Windows

sydvicious:
@whimboo - Please review. This addresses issue #258.

Have the regular expressions for the windows binaries match the ones used for release.

### DIFF
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -782,8 +782,8 @@ class TinderboxScraper(Scraper):
                         'linux64': r'.*\.%(EXT)s$',
                         'mac': r'.*\.%(EXT)s$',
                         'mac64': r'.*\.%(EXT)s$',
-                        'win32': r'.*(\.installer%(STUB)s)\.%(EXT)s$',
-                        'win64': r'.*(\.installer%(STUB)s)\.%(EXT)s$'}
+                        'win32': r'(\.installer%(STUB)s)?\.%(EXT)s$',
+                        'win64': r'(\.installer%(STUB)s)?\.%(EXT)s$'}
 
         regex = regex_base_name + regex_suffix[self.platform]
 


whimboo:
The patch looks fine, but I would appreciate if you could add tests for it. Here my suggestion:

1. For the daily scraper add a test which fetches the zip build of the 2013-10-01 nightly build from mozilla-central
2. For tinderbox also use mozilla-central and use one of the three existent builds. That should be enough
3. For try-builds we have only a single build so it is easy
4. We don't need any tests for candidate and release builds given that we do not have zip archives for those.

Thanks

sydvicious:
@whimboo Note that I added a test for the tinderbox scraper which downloads with --extension=txt. This is much faster than the .zip file, and had the same problem originally. I will add a similar test for the daily scraper, but I don't think I need to download the .zip file. A valid test would be insuring that the .installer.exe file is downloaded without "--extension". Should I do that?

sydvicious:
Does mozdownload support try builds? I get this:

mozdownloadevn)Syds-MBP-2:mozdownload jazzman$ python mozdownload/scraper.py --type=try --branch=mozilla-central --platform=win32
  INFO | Retrieving list of builds from https://ftp.mozilla.org/pub/mozilla.org/firefox/try-builds
  File "mozdownload/scraper.py", line 1260, in <module>
    cli()
  File "mozdownload/scraper.py", line 1252, in cli
    build = BUILD_TYPES[options.type](**kwargs)
  File "mozdownload/scraper.py", line 949, in __init__
  File "mozdownload/scraper.py", line 144, in __init__
    self.get_build_info()
  File "mozdownload/scraper.py", line 954, in get_build_info
    self.builds, self.build_index = self.get_build_info_for_index()
  File "mozdownload/scraper.py", line 1012, in get_build_info_for_index
    raise NotFoundError('No builds have been found', url)
__main__.NotFoundError: No builds have been found: https://ftp.mozilla.org/pub/mozilla.org/firefox/try-builds <https://ftp.mozilla.org/pub/mozilla.org/firefox/try-builds>

Syd Polk - sydpolk@gmail.com
“Let the music be your light.” - Dave Edwards, KUHF-FM, 1982

sydvicious:
Superceded by #263

whimboo:
> Does mozdownload support try builds? I get this:
> 
> mozdownloadevn)Syds-MBP-2:mozdownload jazzman$ python
> mozdownload/scraper.py --type=try --branch=mozilla-central --platform=win32

Yes, it does. But you will have to specify a changeset. Looks like we
have to make it a mandatory argument. Can you please file an issue for
your finding? Thanks.

Issue #252, opened 2015-02-11
Change source of builds from FTP to CDN


Nebelhom:
Looks like there is a source of our builds a bit closer to home

The CDN servers have all our builds in it, as well, and they appear to be faster. Check it out on http://download.cdn.mozilla.net/pub/

Maybe we can try to navigate to those first and if it does not work fall back as a safety to the ftp servers.

whimboo:
@nthomas-mozilla, we have heard rumors that CDN might go away in the future in favor of S3. Would it hurt if we temporarily at least make use of CDN until that happens? We would get some nice speed enhancements that way at least here in Europe.

nthomas-mozilla:
It's fine to use the CDN, and don't worry about how ftp.m.o may change in future (deal with it then!).

Out of curiosity though, do you think it'll help a lot ? Not sure how it will if cache hits are unlikely, as the CDN still needs to talk to ftp.m.o to get a copy. eg I used a shell server in Germany and requested a random nightly directly, and got 5MB/s retrieving 46M, ie 9.1s. Repeating that now that the cache is hot on ftp.m.o gave 7.0 and 12.0s transfers. Using the CDN and a different file took 11.0, 4.6, and 4.7s. ie slightly slower at first but faster on repeated requests.

See also https://bugzilla.mozilla.org/show_bug.cgi?id=1113072

whimboo:
Thanks @nthomas-mozilla for the reply!

Mostly I expect improvements for releases here. I have had a couple of situations in the past months when I tried to download such a build during a release day or so. The download rates are very low. But also on each and every day I can see large drops in the throughput with ftp.m.o in the afternoon my time, which is when US folks are waking up. Download rates of about 700kB/s are usual then.

IMHO we should switch to CDN and we can re-evaluate later if other options exist.

whimboo:
Removing from 1.14 milestone given that we most likely have to wontfix this request. See PR #260.

whimboo:
See PR #260 for the reason why we wontfixed that issue.

PR #260, opened 2015-02-24
Change BASE_URL to CDN servers (#252)

Nebelhom:
This PR address #252 

`BASE_URL` was switched CDN

I tried to create a back_up system in case there are issues with CDN at some point, we automatically fall back to ftp

### DIFF
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -1235,7 +1235,6 @@ def cli():
 
     try:
         build.download()
-
     except KeyboardInterrupt:
         print "\nDownload interrupted by the user"
 
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -36,7 +36,10 @@ mozdownload version: %(version)s
 APPLICATIONS = ('b2g', 'firefox', 'thunderbird')
 
 # Base URL for the path to all builds
-BASE_URL = 'https://ftp.mozilla.org/pub/mozilla.org'
+BASE_URL = 'http://download.cdn.mozilla.net/pub'
+
+# In case CDN fails
+BACKUP_URL = 'https://ftp.mozilla.org/pub/mozilla.org'
 
 # Chunk size when downloading a file
 CHUNK_SIZE = 16 * 1024
@@ -1235,6 +1238,19 @@ def cli():
 
     try:
         build.download()
+
+    except requests.exceptions.ConnectionError as e:
+        if options.url:
+            raise e
+
+        # Fall back to ftp
+        build = BUILD_TYPES[options.type](base_url=BACKUP_URL, **kwargs)
+
+        try:
+            build.download()
+        except KeyboardInterrupt:
+            print "\nDownload interrupted by the user"
+
     except KeyboardInterrupt:
         print "\nDownload interrupted by the user"
 
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -36,10 +36,7 @@ mozdownload version: %(version)s
 APPLICATIONS = ('b2g', 'firefox', 'thunderbird')
 
 # Base URL for the path to all builds
-BASE_URL = 'http://download.cdn.mozilla.net/pub'
-
-# In case CDN fails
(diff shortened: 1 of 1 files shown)


Nebelhom:
@whimboo 

Please review :)

mozdownload/scraper.py
```diff
@@ -36,7 +36,10 @@
 APPLICATIONS = ('b2g', 'firefox', 'thunderbird')
 
 # Base URL for the path to all builds
-BASE_URL = 'https://ftp.mozilla.org/pub/mozilla.org'
+BASE_URL = 'http://download.cdn.mozilla.net/pub'
+
+# In case CDN fails
+BACKUP_URL = 'https://ftp.mozilla.org/pub/mozilla.org'
```
whimboo: I don't think we have to carry the backup URL with us. @nthomas-mozilla what do you say? Is it enough to explicitly work with CDN only? I don't think there will be cases when it will not work.

nthomas-mozilla: AFAIK that CDN URL has everything you need, and it should be reliable.

whimboo: Wonderful. So @Nebelhom please remove the backup logic. We will stay with CDN only, but please make use of HTTPS and not HTTP, which should really be possible.

nthomas-mozilla: I think you'll hit cert issues using https, at least I do loading your BASE_URL in Firefox and getting Akamai. We have an https CDN but it only has installers on it.

whimboo: Ouch, so no way for trustworthy downloads via CDN?

nthomas-mozilla: So https://download-installer.cdn.mozilla.net/pub is everything on ftp except .mar files, and it looks like mozdownload only does 'installers', so just use that domain.

whimboo: We don't only support installer files, but any kind of file which can exist inside a build folder. This included also the packaged tests, the zip build for windows for nightly builds, etc. I had a look at the given URL and was forwarded to ftp.m.o in case for nightly, tinderbox, and try builds. I can see all the files we need even including the mar files. So I'm not sure what you meant that mar files will not be present. Does your comment only apply to candidates and releases? Otherwise it looks good and good to see that there is a way to retrieve files via HTTPS.

Nick what I wonder is if connections are going outside of ftp.m.o to the CDN network for at least the release builds. When I navigate through the folder structure and trying to download various installers, always ftp.mo is used as backend server.

Nebelhom:
@whimboo 

I removed the back_up and changed the `BASE_URL` to https://download-installer.cdn.mozilla.net/pub . That should be it.

Please review :)

mozdownload/scraper.py
nthomas-mozilla: You weren't actually forwarded to ftp.m.o in the sense of a 302. The headers you get from a request to the CDN are sum of
* the request from CDN to origin server
* your request to the CDN
The origin server is a vhost on the same cluster that serves ftp.m.o, so you get this sort of thing
  X-Backend-Server: ftp5.dmz.scl3.mozilla.com
when making a CDN request, and it doesn't mean you talked to ftp.m.o directly.

The origin server for https://download-installer.cdn.mozilla.net blocks downloads of .mar files at the time of download, but doesn't modify directory listings. eg if you try to download
 https://download-installer.cdn.mozilla.net/pub/firefox/releases/36.0/update/win32/en-US/firefox-34.0.5-36.0.partial.mar
you get a 403.

whimboo:
Given the limitation that we are not able to allow the download of .mar files, we should not switch away from FTP. I'm not sure if we have customers who actually rely on those files, and I don't want to introduce a regression.

Looks like we have to wontfix this feature request, and have to wait until we get something on S3. @nthomas-mozilla do you agree?

whimboo:
Sorry Johannes for the wasted time on this PR but I still vote for wontfix. It looks like Nick also doesn't have another option for us. Closing as wontfix.

Nebelhom:
@whimboo Don't worry. The aim is to improve mozdownload. If it does not improve it, then what's the point? ;)

Issue #257, opened 2015-02-14
Test: Correct Choice of Scraper


Nebelhom:
This are tests that still needs automation and are inter-related.

```
mozdownload -t release -p win32 -v latest
mozdownload -t candidate -p win32 -v 21.0
mozdownload -t daily -p win32
mozdownload -t tinderbox -p win32
```

Nebelhom:
@whimboo @davehunt and anyone else knowledgeable in writing tests.

I have been pondering over this particular implementation for quite some time, but cannot find a good solution of it.

Essentially, when implementing this, we need to call the `cli()` function and then check if we have the correct scraper. Anything else, in my eyes, is not really testing but writing a function that gives you a pre-determined answer at point of writing.

At the moment, `cli()` is a closed sequence, that when called, checks the arguments passed on, chooses a scraper accordingly and then starts downloading a build if possible.

The only way I have found to create a decent test for it, is to break up `cli()`into two parts. The "matching-the-arguments" part and the "downloading" part (Passing arguments in_between the two)

Q1: Can anyone think of a better, more elegant or simpler way?

Q2: Is it actually feasible to break up the `cli()` function, to allow for more effective testing? We don't have to implement test-driven development, but we are going in the direction of having tests in place for each function, anyways...

Looking forward to a constructive discussion :)

whimboo:
Personally I would say that we have perfect coverage for that already. I don't think we have to invest our time for checking if the correct scraper has been used. Reason is that if a wrong scraper would have been chosen, other existing tests would fail. There might only be a small gap here between release and candidate scrapers. But even there we have to `buildX` part we expect for the latter.

So lets drop it?

Nebelhom:
@whimboo 

I think you forget that we never test if the scraper chooses the right type. For our tests, we directly invoke the correct type for those tests. Check the following link:

https://github.com/mozilla/mozdownload/blob/master/tests/daily_scraper/test_daily_scraper.py#l318

So as you can see, for all the tests that we are doing, we directly create an instance of `DailyScraper`, rather than letting the script come to its own conclusion that we need the DailyScraper.

I agree that we do not want to do it, if we only check for the correct choice of scraper, because we would only check for typos in the `BUILD-TYPE` block ( https://github.com/mozilla/mozdownload/blob/master/tests/daily_scraper/test_daily_scraper.py#l318 ). However, it might be worthwhile to write tests for all flags. Typos are easily done and they are hard to spot once they are in.

[...]

PR #270, opened 2015-03-12
Test for correct choice of scraper (#257)

Nebelhom:
This PR addresses issue #257 

In order to make this test work, I added the `--base_url` flag, which allows you to choose the base_url of your server. I could not think of another way of calling scraper and using mozhttp.

### DIFF
--- a/tests/base_scraper/test_correct_scraper.py
+++ b/tests/base_scraper/test_correct_scraper.py
@@ -1,0 +1,70 @@
+#!/usr/bin/env python
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import os
+import unittest
+import urllib
+
+import mozfile
+import mozhttpd_base_test as mhttpd
+from mozprocess import processhandler
+
+
+class CorrectScraperTest(mhttpd.MozHttpdBaseTest):
+    """Test mozdownload for correct choice of scraper"""
+
+    def test_scraper(self):
+        """Testing various download scenarios for DailyScraper"""
+
+        # This works
+        # Now choose four distinctly named packages
+        # Download them and check if correct file name
+        # Then delete again for each cycle
+
+        # Once whole test is written, fix manifest.ini's again.
+
+        tests = [
+            # ReleaseScraper
+            {'command': ['./mozdownload/scraper.py', '-v', 'latest',
+                         '--base_url=%s' % self.wdir,
+                         '--destination=%s' % self.temp_dir],
+             'fname': 'firefox-latest.en-US.linux64.tar.bz2'},
+            # ReleaseCandidateScraper
+            {'command': ['./mozdownload/scraper.py', '-t', 'candidate', '-v',
+                         '21.0', '-p', 'win32',
+                         '--base_url=%s' % self.wdir,
+                         '--destination=%s' % self.temp_dir],
+             'fname': 'firefox-21.0-build3.en-US.win32.exe'},
+            # DailyScraper
+            {'command': ['./mozdownload/scraper.py', '-t', 'daily',
+                         '-p', 'win32', '--base_url=%s' % self.wdir,
+                         '--destination=%s' % self.temp_dir],
+             'fname': '2013-10-01-03-02-04-mozilla-central-firefox-27.0a1.en-US.win32.installer.exe'},
+            # TinderboxScraper
+            {'command': ['./mozdownload/scraper.py', '-t', 'tinderbox',
+                         '-p', 'win32', '--base_url=%s' % self.wdir,
+                         '--destination=%s' % self.temp_dir],
+             'fname': 'mozilla-central-firefox-25.0a1.en-US.win32.installer.exe'},
+            # TryScraper
+            {'command': ['./mozdownload/scraper.py', '-t', 'try', '-p',
+                         'mac64', '--changeset=8fcac92cfcad',
+                         '--base_url=%s' % self.wdir,
+                         '--destination=%s' % self.temp_dir],
+             'fname': '8fcac92cfcad-firefox-38.0a1.en-US.mac.dmg'},
+        ]
+
+        for entry in tests:
+            p = processhandler.ProcessHandler(entry['command'])
+            p.run()
+            p.wait()
+            dir_content = os.listdir(self.temp_dir)
+            self.assertTrue(entry['fname'] in dir_content)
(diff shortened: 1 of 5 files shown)


Nebelhom:
@whimboo 

Could you give me a review please? Thanks a lot and enjoy your weekend! :)

sydvicious:
@whimboo is out for a while. Given your discussion, the clean test results, and the rather straightfoward patch, I am giving a r+

Nebelhom:
@sydvicious Thanks for the review! Sorry for the late reply, I was having issues with my PC the last two weeks that did not allow me to go online.

Do you know, how long @whimboo will be away. We talked about this test in IRC at length and I am not sure whether this fix may break some of the other uses of mozdownload. @whimboo has a good overview in that department.

tests/base_scraper/test_correct_scraper.py
```diff
@@ -0,0 +1,70 @@
...
+        """Testing various download scenarios for DailyScraper"""
+
+        # This works
+        # Now choose four distinctly named packages
+        # Download them and check if correct file name
+        # Then delete again for each cycle
+
+        # Once whole test is written, fix manifest.ini's again.
+
+        tests = [
```
whimboo: In all other tests we defined the data globally. I would suggest that we follow that rule and do it similar here. Also I think the comment above is not necessary anymore, right?

tests/base_scraper/test_correct_scraper.py
```diff
@@ -0,0 +1,70 @@
...
+            # ReleaseScraper
+            {'command': ['./mozdownload/scraper.py', '-v', 'latest',
+                         '--base_url=%s' % self.wdir,
+                         '--destination=%s' % self.temp_dir],
```
whimboo: The command, base url, and destination is always the same. I would hardcode it in the test, and extend the arguments as necessary.

tests/base_scraper/test_correct_scraper.py
```diff
@@ -0,0 +1,70 @@
...
+
+        for entry in tests:
+            p = processhandler.ProcessHandler(entry['command'])
+            p.run()
+            p.wait()
+            dir_content = os.listdir(self.temp_dir)
+            self.assertTrue(entry['fname'] in dir_content)
+
+            if entry['fname'] in dir_content:
+                mozfile.remove(os.path.join(self.temp_dir, entry['fname']))
```
whimboo: There is no need for the if condition here. We only reach this code when the assert above does not fail.

tests/test_cases.md
```diff
@@ -1,13 +1,5 @@
 # Test Cases
 
-## Correct Choice of Scraper
-```
-mozdownload -t release -p win32 -v latest
-mozdownload -t candidate -p win32 -v 21.0
-mozdownload -t daily -p win32
-mozdownload -t tinderbox -p win32
-```
-
```
whimboo: hurray!

whimboo:
Beside the nits all look fine. Lets fix that and we can get it landed. Thanks!

tests/base_scraper/test_correct_scraper.py
Nebelhom: I did it, because otherwise I would not have `self.wdir` and `self.temp_dir` available.

Your next nitpik below, however, takes care of that ;)

Nebelhom:
@whimboo Thanks for the feedback.

Rather nice timing, that we both come back at the same time from a long-ish layoff ;)

tests/base_scraper/test_correct_scraper.py
```diff
@@ -0,0 +1,58 @@
...
+
+import mozfile
+import mozhttpd_base_test as mhttpd
+from mozprocess import processhandler
+
+tests = [
+    # ReleaseScraper
+    {'command': ['-v', 'latest'],
+     'fname': 'firefox-latest.en-US.linux64.tar.bz2'},
+    # ReleaseCandidateScraper
```
whimboo: nit: lets leave blank lines between the examples so its easier to distinguish. I think we do the same for the other tests.

tests/base_scraper/test_correct_scraper.py
```diff
@@ -0,0 +1,58 @@
...
+    {'command': ['-t', 'candidate', '-v',
+                 '21.0', '-p', 'win32',],
```
whimboo: When you break lines, please do in such a way that we do not separate options which have to be used together.

tests/base_scraper/test_correct_scraper.py
```diff
@@ -0,0 +1,58 @@
...
+
+class CorrectScraperTest(mhttpd.MozHttpdBaseTest):
+    """Test mozdownload for correct choice of scraper"""
+
+    def test_scraper(self):
+        """Testing various download scenarios for DailyScraper"""
+
+        for entry in tests:
+            #p = processhandler.ProcessHandler(entry['command'])
+            commands = ['./mozdownload/scraper.py',
```
whimboo: The above commented out line can be removed. Also this variable I would call `command` because you do not have more then that. The other two are just options.

Nebelhom:
@whimboo Please review. Thanks for the review :)

tests/base_scraper/test_correct_scraper.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env python
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import os
+import unittest
+import urllib
```
whimboo: just seen that urllib is not used.

tests/base_scraper/test_correct_scraper.py
```diff
@@ -0,0 +1,57 @@
...
+
+import mozfile
+import mozhttpd_base_test as mhttpd
+from mozprocess import processhandler
+
+tests = [
+    # ReleaseScraper
+    {'command': ['-v', 'latest'],
```
whimboo: maybe calling it options, which is a better fit.

Nebelhom:
@whimboo and the next round ;)

whimboo:
Looks fine to me now. If Travis also passes we can get it merged.

Nebelhom:
Awesome! Let's wait for it and see.

My first personal merger for mozdownload :) I got "mozjitters" :D

whimboo:
The merge was a bit problematic but basically this PR got merged as:
https://github.com/mozilla/mozdownload/commit/37e8ae1672cfb3e90d1c54bdd240400eb13204e7

PR #272, opened 2015-03-25
Connection close

sydvicious:
When we traverse directories in ftp.mozilla.org, we don't close the http connection. In urllib, at least on some platforms, when you close a connection, it will be reused. Without this, it is easy to flood ftp.mozilla.org with connections, and if you are not on Mozilla's network, ftp.mozilla.org will start returning 503 Server Too Busy errors.

### DIFF
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -431,7 +431,9 @@ class DailyScraper(Scraper):
                          auth=self.authentication, headers=headers)
         r.raise_for_status()
 
-        return datetime.strptime(r.text.split('\n')[0], '%Y%m%d%H%M%S')
+        retval = datetime.strptime(r.text.split('\n')[0], '%Y%m%d%H%M%S')
+        r.close()
+        return relval
 
     def is_build_dir(self, dir):
         """Return whether or not the given dir contains a build."""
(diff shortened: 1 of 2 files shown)


AutomatedTester:
r+

sydvicious:
From IRCE: AutomatedTester|busy
11:22 sydpolk: r+, ship it

whimboo:
The last commit on this PR let Travis fail and should not have been pushed to master! @sydvicious, I will send you instructions in how to merge PRs via a separate email. Please follow those items in the future. Thanks.

mozdownload/scraper.py
```diff
@@ -431,7 +431,9 @@ def get_latest_build_date(self):
                          auth=self.authentication, headers=headers)
         r.raise_for_status()
 
-        return datetime.strptime(r.text.split('\n')[0], '%Y%m%d%H%M%S')
+        retval = datetime.strptime(r.text.split('\n')[0], '%Y%m%d%H%M%S')
+        r.close()
+        return relval
```
whimboo: This will still not work in case of exceptions before the `close()` call. A better approach here would have been a try/finally block. Same for the other replacement.

PR #275, opened 2015-03-30
Wrap the work with an http connection in a try-finally block so we will always close it.

sydvicious:


sydvicious:
@whimboo - Please review. Thanks.

whimboo:
So this is a follow-up for PR #272, which was not complete. And we want this for the 1.15 release.

mozdownload/scraper.py
```diff
@@ -427,12 +427,17 @@ def get_latest_build_date(self):
...
         headers = {'Cache-Control': 'max-age=0'}
+
+        retval = None
         r = requests.get(url + parser.entries[-1],
                          auth=self.authentication, headers=headers)
-        r.raise_for_status()
+        try:
+            r.raise_for_status()
+
+            retval = datetime.strptime(r.text.split('\n')[0], '%Y%m%d%H%M%S')
```
whimboo: You can return immediately. There is no need for the extra retval variable. In case of an exception the retval is not used at all.

sydvicious:
@whimboo Please review latest change. Thanks.

mozdownload/scraper.py
whimboo: As I told in my last review the `retval` variable is not necessary at all. Please remove its declaration and any usage of it. In this line you can return directly.

mozdownload/scraper.py
```diff
@@ -427,12 +427,17 @@ def get_latest_build_date(self):
...
+        finally:
+            r.close()
 
-        retval = datetime.strptime(r.text.split('\n')[0], '%Y%m%d%H%M%S')
-        r.close()
         return retval
```
whimboo: Please also remove that line.

sydvicious: @whimboo: So the latest commit in my branch in my fork has this removed. How do I update the pull request?

sydvicious:
@whimboo Set the git upstream branch correctly, so latest change is pushed. Please review. Thanks.

whimboo:
That looks perfect now. Let me merge it. Thanks for noticing that heavy glitch!

whimboo:
This PR got merged as https://github.com/mozilla/mozdownload/commit/13a0860caf3a964f951581488f98aa879221f626

PR #277, opened 2015-05-02
Upgrade dependency for requests module to 2.7.0 (#271)

### LINKED ISSUE #271: Upgrade dependency for requests module to 2.6.0
We are still using requests v1.2.0 which is kinda outdated. We should upgrade to the latest available version 2.6.0: https://pypi.python.org/pypi/requests

TLemur:
Just changed version number and run tests. All went smoothly.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ version = '1.14'
 deps = ['mozinfo >= 0.7',
         'mozlog >= 1.3',
         'progressbar == 2.2',
-        'requests >= 2.7.0',
+        'requests == 2.7.0',
         ]
 
 setup(name='mozdownload',
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ version = '1.14'
 deps = ['mozinfo >= 0.7',
         'mozlog >= 1.3',
         'progressbar == 2.2',
-        'requests >= 2.5.0',
+        'requests >= 2.7.0',
         ]
 
 setup(name='mozdownload',
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ version = '1.14'
 deps = ['mozinfo >= 0.7',
         'mozlog >= 1.3',
         'progressbar == 2.2',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -19,7 +19,7 @@
 deps = ['mozinfo >= 0.7',
         'mozlog >= 1.3',
         'progressbar == 2.2',
-        'requests == 1.2.2',
+        'requests >= 2.5.0',
```
whimboo: Version 2.7.0 has been released 3 days ago. We should upgrade to that one.

whimboo:
Btw. this PR will fix issue #271.

mnoorenberghe:
Any reason for the delay? mozdownload doesn't work for me without this (I think it's something to do with the certificate validation).

whimboo:
Looks like since my last review a new commit has been pushed to this PR. Given that Github does not send out any email notification for that it was simply missed by myself. I will have a look so we can get this landed.

@mnoorenberghe not sure why it is not working for you without this change. We do not experience a problem with mozdownload 1.14 in our CI system, and no-one else complained about it yet. :S Lets see that we can get a new version out soon.

setup.py
```diff
@@ -19,7 +19,7 @@
 deps = ['mozinfo >= 0.7',
         'mozlog >= 1.3',
         'progressbar == 2.2',
-        'requests == 1.2.2',
+        'requests >= 2.7.0',
```
whimboo: For packages outside of our control we do not want this non-strict version matches. Please keep the double equal signs as what we had before. Thanks. Once you update the PR please also add a comment. I will land this patch then.

TLemur:
>I will land this patch then.
Can you do this please?

whimboo:
Thanks @TLemur! It looks fine now and I merged your PR into master as https://github.com/mozilla/mozdownload/commit/0c22ad25faedc68ef7c71703c54cd07e513dc5a3. Let us know if you want to work on something else. There are a couple other mentored issues available.

Issue #67, opened 2013-04-11
Fallback to win32 builds on win64


whimboo:
On all Windows 64bit machines we currently fail to download the build because it defaults to win64 builds which we currently don't ship. Lets default to win32 builds instead.

whimboo:
Long time no update here. Soon we will ship win64 builds. Right now those are on beta:
ftp://ftp.mozilla.org/pub/firefox/releases/latest-beta/win64/

So I don't think that we should fix this issue anymore. It's simply outdated. @TLemur sorry for not updating it in time and that you wasted time for trying to fix it. :(

TLemur:
@whimboo No problem, just can you answer my question in issue #108 please?

PR #281, opened 2015-05-30
Falls back to win32 builds on win64 platform (#67)

TLemur:
This PR should fix issue #67


whimboo:
Please see issue #67 why this will not be an issue anymore soon. I'm going to close it as wontfix.

Issue #282, opened 2015-06-01
Release version 1.15


whimboo:
We want to release version 1.15 of mozdownload.

The following tasks have to be performed:
- [x] Update setup.py for version bump
- [x] Update changelog
- [ ] Tag version 1.15
- [ ] Release to pypi

whimboo:
Version 1.15 has been tagged:
https://github.com/mozilla/mozdownload/commits/1.15

Pushed to pypi:
Submitting dist/mozdownload-1.15.tar.gz to https://pypi.python.org/pypi
Server response (200): OK

PR #284, opened 2015-06-01
Release version 1.15 (#282)

whimboo:
This PR fixes issue #282. @sydvicious if you can review please? Thanks. Maybe also test if all is working fine for your specific cases, so we can be sure we indeed fixed all the failures.


AutomatedTester:
r+

Issue #250, opened 2015-02-11
Stop running Travis CI tests for Python 2.6


whimboo:
All of our systems have been upgraded to Python 2.7, so I do not see any need to let Travis CI run tests via Python 2.6. This version of Python should not be used anymore. So lets get it removed.

whimboo:
Fixed with PR #296.

PR #287, opened 2015-06-04
Various Travis CI and related issues (closes #157, #250)

### LINKED ISSUE #157: Enhance Travis CI job with more validity checks
Lets consider adding pyflake to our travis ci job description. That would help us even more to have sane code in our repository.

TLemur:
Please merge, it's ready now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
-  - "2.6"
   - "2.7"
 install:
   - pip install pep8
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ python:
 install:
   - pip install pylama
 before_script:
-  - "pylama mozdownload"
+  - "pylama -i C901 mozdownload"
 script: ./run_tests.py
 notifications:
   email:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,9 @@ language: python
 python:
   - "2.7"
 install:
-  - pip install pep8
+  - pip install pylama
 before_script:
-  - "pep8 mozdownload"
+  - "pylama mozdownload"
 script: ./run_tests.py
 notifications:
   email:
--- a/mozdownload/__init__.py
+++ b/mozdownload/__init__.py
@@ -2,4 +2,4 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from scraper import *
+from scraper import * #noqa
--- a/mozdownload/__init__.py
+++ b/mozdownload/__init__.py
@@ -2,4 +2,15 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from scraper import * #noqa
+from scraper import PLATFORM_FRAGMENTS, NotSupportedError, NotFoundError, \
+    NotImplementedError, TimeoutError, Scraper, DailyScraper, \
+    DirectScraper, ReleaseScraper, ReleaseCandidateScraper, \
+    TinderboxScraper, TryScraper
+from parser import DirectoryParser
+from timezones import PacificTimezone
+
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -1,11 +1,10 @@
 language: python
 python:
-  - "2.6"
   - "2.7"
 install:
-  - pip install pep8
+  - pip install pylama
```
whimboo: Nice one! I haven't known about this tool. It looks great that it wraps a couple of tools into a single one.

.travis.yml
```diff
@@ -1,11 +1,10 @@
 language: python
 python:
-  - "2.6"
   - "2.7"
```
whimboo: So this fixes issue #250.

mozdownload/__init__.py
```diff
@@ -2,4 +2,4 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from scraper import *
+from scraper import * #noqa
```
whimboo: Please get this comment removed.

TLemur: If I remove the comment, it won't pass one of pylama's check

.travis.yml
TLemur: Both #250 and #157

mozdownload/__init__.py
whimboo: I assume it's because of the `import *`? Maybe we should simply state every supported scraper class.

TLemur: I tried this as well. But then pylama complains they are not used.

whimboo: Given that wild-card imports are considered bad, you can import each scraper class individually and add them to a `__all__` array.

TLemur:
Given that it is passing Travis CI, I don't doubt it can be merged now.

mozdownload/__init__.py
```diff
@@ -2,4 +2,15 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from scraper import *
+from scraper import PLATFORM_FRAGMENTS, NotSupportedError, NotFoundError, \
+    NotImplementedError, TimeoutError, Scraper, DailyScraper, \
+    DirectScraper, ReleaseScraper, ReleaseCandidateScraper, \
+    TinderboxScraper, TryScraper
```
whimboo: I would suggest that we limit the number of exported symbols to what external code really needs. That means please remove `PLATFORM_FRAGMENTS` which is used only internally. For the others I would suggest to use one per line, and all surrounded by brackets, so you can avoid the backslashes. See the following link as an example: https://github.com/mozilla/firefox-ui-tests/blob/master/firefox_ui_harness/runtests.py#L13

mozdownload/__init__.py
```diff
@@ -2,4 +2,15 @@
...
+from parser import DirectoryParser
+from timezones import PacificTimezone
```
whimboo: As a normal user you don't need those. If necessary they need to be imported when needed.

mozdownload/__init__.py
TLemur: I agree that as a normal user you do not need PLATFORM_FRAGMENTS, but tests fail is it's removed.

mozdownload/__init__.py
TLemur: Again, tests fail.

mozdownload/__init__.py
whimboo: Tests should import this constant via `from mozdownload.scraper import PLATFORM_FRAGMENTS`. That will still be allowed.

mozdownload/__init__.py
whimboo: Maybe you are seeing #293? It's a perma fail for me locally.

whimboo:
I went ahead and fixed the remaining issues in PR #296. So I'm going to close this out.

PR #289, opened 2015-06-06
Removed all unused variables from get_build_info_for_version (fixes #169)

### LINKED ISSUE #169: version argument in get_build_info_for_version unused
I noticed that in `get_build_info_for_version` the `version` argument is never used.

https://github.com/mozilla/mozdownload/blob/master/mozdownload/scraper.py#l576

Is this an artefact of a previous fix? Can we remove it (and fix all calls to this method)?

TLemur:
Removed version which is not used at all and build_index which is always None. Fixes issue #169.

### DIFF
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -657,8 +657,7 @@ class ReleaseCandidateScraper(ReleaseScraper):
         "Defines additional build information"
 
         # Internally we access builds via index
-        self.builds, self.build_index = self.get_build_info_for_version(
-            self.version)
+        self.builds, self.build_index = self.get_build_info_for_version()
         if self.build_number and \
                 ('build%s' % self.build_number) in self.builds:
             self.builds = ['build%s' % self.build_number]
@@ -668,7 +667,7 @@ class ReleaseCandidateScraper(ReleaseScraper):
             self.logger.info('Selected build: build%d' %
                              (self.build_index + 1))
 
-    def get_build_info_for_version(self, version, build_index=None):
+    def get_build_info_for_version(self):
         url = urljoin(self.base_url, self.candidate_build_list_regex)
 
         self.logger.info('Retrieving list of candidate builds from %s' % url)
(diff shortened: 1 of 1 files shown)


mozdownload/scraper.py
```diff
@@ -657,8 +657,7 @@ def get_build_info(self):
         "Defines additional build information"
 
         # Internally we access builds via index
-        self.builds, self.build_index = self.get_build_info_for_version(
-            self.version)
+        self.builds, self.build_index = self.get_build_info_for_version()
```
whimboo: `self.version` is not coming through as function parameter but it is a class member. So I don't see why we can remove it here, especially when the underlying method expects an explicit version number.

mozdownload/scraper.py
```diff
@@ -668,7 +667,7 @@ def get_build_info(self):
             self.logger.info('Selected build: build%d' %
                              (self.build_index + 1))
 
-    def get_build_info_for_version(self, version, build_index=None):
+    def get_build_info_for_version(self):
         url = urljoin(self.base_url, self.candidate_build_list_regex)
```
whimboo: Hm, now I see why you are removing those parameters. The version is directly used as `self.version` in `candidate_build_list_regex`, so passing it through doesn't make sense. While looking at especially this method I would say lets kill it and integrate its code to `get_build_info()`.

TLemur:
Ok, I shall integrate its code to get_build_info().

whimboo:
This code looks fine to me now. @TLemur can you please squash your commits and set a proper commit message with the issue id in bracktes at the end? See other changesets how it has to look like. Let me know once done so I can land this change. Thanks!

TLemur:
@whimboo what do you mean by "squash"?

whimboo:
Right now this PR contains two commits. To check it in, it has to be a single one. So you have to interactively rebase the commits. See the [docs](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) for help. For the selection use 'f' for each commit beside the first one, which you want to mark as 'r' to rename the message.

AutomatedTester:
landed in 655b3e4

PR #291, opened 2015-06-08
Add waffle.io badge to README

whimboo:


whimboo:
@Nebelhom could you have a look at it? I think it will help us to better schedule issues.

Nebelhom:
@whimboo I do not see anything wrong with it.

Nice tool, but what triggered you to use it?

whimboo:
A blog post from a colleague. I thought that this project might be a good idea to at least get it tested. If it works I would consider it for other projects too, if not we can get rid of it at anytime.

Issue #112, opened 2013-07-23
Add support for downloading Fennec (Firefox Mobile) builds


davehunt:
Releases: https://ftp.mozilla.org/pub/mozilla.org/mobile/releases/
Candidates: https://ftp.mozilla.org/pub/mozilla.org/mobile/candidates/
Tinderbox: https://ftp.mozilla.org/pub/mozilla.org/mobile/tinderbox-builds/
Daily: https://ftp.mozilla.org/pub/mozilla.org/mobile/nightly/

ncalexan:
I would like to help mentor this as well.  This could be part of https://bugzilla.mozilla.org/show_bug.cgi?id=1124378.

Nebelhom:
@davehunt @ncalexan 

Is fennec's directory structure similar to one of the builds we already use (e.g. B2G or firefox)? I am just curious if it is a whole portion of code or more of a small change to a couple of lines of code...

davehunt:
They certainly appear to be similar, but I'm not familiar enough to say with confidence.

ncalexan:
Yes, similar if not identical.  At the end of the day, the packaging code
is the same -- `packager.{mk,py}` and `upload-files.mk`.  There tweaks to
the filename but the locale hierarchy, package name root, etc should follow
the pattern.

whimboo:
Not sure if it is helpful right now to spend time on this issue. We are going to have to move to S3 for downloading builds soon. See issue #278. Maybe we can get this for free then without having to code that much as we would have to do now.

dminor:
I'm planning to migrate mozbench to using mozdownload and need fennec support for that, so I'm planning to work on this.

dminor:
Since mozbench only needs 'daily' builds, I'm only going to add support for that, rather than put in a bunch on effort that may have been unnecessary after we move to S3.

whimboo:
That sounds totally fine. Thanks Daniel!

whimboo:
Via PR #292 we got support for downloading Fennec daily builds. I will keep this issue open for the remaining build types. Thanks Dan for the new feature!

whimboo:
Fenix support has actually been added via https://github.com/mozilla/mozdownload/issues/624 and is available in the latest mozdownload release. Marking as dupe.

PR #292, opened 2015-06-12
Add support for downloading Fennec (Firefox Mobile) daily builds (#112)

dminor:
This adds support for daily builds for the android-api-9, android-api-11
and android-x86 platforms. Supporting other types of builds may be easier
when things move from ftp to S3 so that work is deferred until then.

### DIFF
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -33,7 +33,7 @@ applications.
 mozdownload version: %(version)s
 """ % {'version': version}
 
-APPLICATIONS = ('b2g', 'firefox', 'thunderbird')
+APPLICATIONS = ('b2g', 'firefox', 'mobile', 'thunderbird')
 
 # Base URL for the path to all builds
 BASE_URL = 'https://ftp.mozilla.org/pub/mozilla.org'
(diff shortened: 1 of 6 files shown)


whimboo:
@dminor, you missed to add the information to a comment that this PR fixes issue #112, also due to a missing cc I totally missed this PR.

So is this PR ready for a review?

dminor:
@whimboo yes, please review it :)

mozdownload/scraper.py
```diff
@@ -414,7 +420,11 @@ def get_build_info(self):
 
     def get_latest_build_date(self):
         """ Returns date of latest available nightly build."""
-        url = urljoin(self.base_url, 'nightly', 'latest-%s/' % self.branch)
+        if not self.application == 'mobile':
```
whimboo: I would suggest we directly use `not in ('mobile')` here so we are prepared for other applications to get added.

mozdownload/scraper.py
```diff
@@ -469,11 +479,14 @@ def get_build_info_for_date(self, date, build_index=None):
         self.logger.info('Retrieving list of builds from %s' % url)
         parser = DirectoryParser(url, authentication=self.authentication,
                                  timeout=self.timeout_network)
-        regex = r'%(DATE)s-(\d+-)+%(BRANCH)s%(L10N)s$' % {
+        is_mobile = self.application == 'mobile'
```
whimboo: I would move it inline to line 488 and also make use of `not in ('mobile')`, so default comes first as for l10n selection.

mozdownload/scraper.py
```diff
@@ -516,7 +532,8 @@ def binary_regex(self):
                         'win64': r'(\.installer%(STUB)s)?\.%(EXT)s$'}
         regex = regex_base_name + regex_suffix[self.platform]
 
-        return regex % {'APP': self.application,
+        is_mobile = self.application == 'mobile'
+        return regex % {'APP': 'fennec' if is_mobile else self.application,
```
whimboo: Same here for `not in`. Also I tend to think a mapping would make sense to map from an application name to the folder name on FTP. I would rather have `fennec` as application name than `mobile` given that this is the real application name.

whimboo:
Great patch and I would like to see this included in the next release of mozdownload. @dminor if that feature is important for you lets try to get it landed at latest on Monday. Then I will have a chance to release a new version before I head out for a full month. Do you think that would be possible? I think so given that it's mostly only nits to solve.

dminor:
@whimboo thanks for the feedback. I would like to get this landed before you go. I'll check in with you Monday morning assuming you don't see this before you're done work today.

whimboo:
I like those changes! Looks like you addressed them all. I will wait for the travis results and do some tests myself. But all in all this looks to be ready for being merged. I will do the merge once my testing doesn't show any issues. Thanks!

whimboo:
I squashed the commits and renamed the commit message to also include the PR #.

PR got merged as https://github.com/mozilla/mozdownload/commit/41f50ade4dd2fde381774599063e972cec093cac.

I will release a new version in a bit. Thanks for contributing this feature Dan!

whimboo:
@dminor, this feature is available now in the newly released version 1.16 of mozdownload.

Issue #293, opened 2015-06-26
Update dependencies for python packages for test environment


whimboo:
There is a problem in installing moznetwork due to a conflict with mozlog.structured. As best we should bump up all dependencies and don't make them strict. The strict mode is not necessary for running the tests, for the product itself I would keep it for now.

whimboo:
PR #294 has been merged.

PR #294, opened 2015-06-26
Update dependencies for python packages for test environment (#293)

whimboo:
This fixes issue #293, and allows to run the tests again. When Travis shows passing tests, I will get it merged immediately.


whimboo:
The failure as shown here happens with Python 2.6, which we want to get rid of anyway. So I will get this PR merged now.

PR #298, opened 2015-07-06
Add scraper.cli to __init__.py

dminor:
This fixes what appears to be the accidental removal of scraper.cli from the module __init__.py which prevents calling the module directly to perform downloads. This also adds a test to ensure that mozdownload is callable from the shell.

### DIFF
--- a/mozdownload/__init__.py
+++ b/mozdownload/__init__.py
@@ -9,6 +9,7 @@ from .scraper import (Scraper,
                       ReleaseCandidateScraper,
                       TinderboxScraper,
                       TryScraper,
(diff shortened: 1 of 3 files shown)


AutomatedTester:
r+

whimboo:
Thanks @dminor for the fix. Someone should have released version 1.17 with it. I see it somewhat critical to get out. I will do so later today.

whimboo:
@dminor, just a hint for you.  When you merge PRs please do it via the command line tools and squash all the commits first. Only that will give us a clean history. Thanks.

Issue #299, opened 2015-07-23
AttributeError: 'module' object has no attribute 'getLogger'


crdoconnor:
This error occurs on the latest (master) version of mozdownload::

    Traceback (most recent call last):
      File "/home/user/.local/bin/mozdownload", line 9, in <module>
        load_entry_point('mozdownload==1.16', 'console_scripts', 'mozdownload')()
      File "/home/user/.local/venvs/mozdownload/local/lib/python2.7/site-packages/mozdownload/scraper.py", line 1248, in cli
        build = BUILD_TYPES[options.type](**kwargs)
      File "/home/user/.local/venvs/mozdownload/local/lib/python2.7/site-packages/mozdownload/scraper.py", line 592, in __init__
        Scraper.__init__(self, *args, **kwargs)
      File "/home/user/.local/venvs/mozdownload/local/lib/python2.7/site-packages/mozdownload/scraper.py", line 107, in __init__
        self.logger = mozlog.getLogger(' ')
     AttributeError: 'module' object has no attribute 'getLogger'

It looks like the latest version of mozlog no longer supports the API mozdownload is trying to use.

whimboo:
PR #300 actually fixed the breakage with `getLogger()` due to API changes. But I still wonder to get rid of mozlog and simply use the `logging` module. I will have a look into it.

whimboo:
I will push the follow-up through another issue. Marking that one as fixed.

PR #300, opened 2015-07-27
Adapt to mozlog 3.0 (bug 1014760)

jonallengriffin:


jonallengriffin:
Passing to @AutomatedTester for review.

AutomatedTester:
r+

AutomatedTester:
landed in a99d85b

whimboo:
Thank you for the fix @jonallengriffin! This actually fixes the reported issue #299.

But I wonder why we are making use of mozlog at all now given that we only use it for `getLogger()`. Shouldn't we simply use the logging module?

@jonallengriffin, also please do not create new branches on the upstream repository but on your fork instead. Thanks.

jonallengriffin:
mozlog's getLogger does a little more work for you than the base Python method, so it doesn't hurt to keep it IMO

whimboo:
It's just an extra line for calling `basicConfig()` with the wanted formatting setting. I don't think it qualifies to depend on mozlog. I will fix it in a follow-up issue/pr.

Issue #316, opened 2015-09-10
Move CLI into its own module


whimboo:
Having the CLI code in the scraper doesn't make sense and only blows up its size. Lets separate it out into its own module.

whimboo:
PR #317 got merged.

PR #317, opened 2015-09-10
Move CLI into its own module (#316)

whimboo:
This PR fixes issue #316. @davehunt mind reviewing it?


mozdownload/cli.py
```diff
@@ -0,0 +1,240 @@
...
+import sys
+
+from . import errors
+from . import scraper
+
+
+version = pkg_resources.require("mozdownload")[0].version
+
+__doc__ = """
+Module to handle downloads for different types of archive.mozilla.org hosted \
```
AutomatedTester: the `\` isnt needed

mozdownload/cli.py
```diff
@@ -0,0 +1,240 @@
...
+    parser.add_option('--url',
+                      dest='url',
+                      metavar='URL',
+                      help='URL to download. Note: Reserved characters (such '
+                           'as &) must be escaped or put in quotes otherwise '
+                           'CLI output may be abnormal.')
+    parser.add_option('--version', '-v',
+                      dest='version',
+                      metavar='VERSION',
+                      help='Version of the application to be used by release\
```
AutomatedTester: You have used `\` and silent concat (e.g. `"foo" "bar"` == "foobar"`). Please try be consistent.

mozdownload/cli.py
whimboo: That's an old relict and hasn't been newly added.  Would you be ok when I fix that with my next PR? It's already fixed there locally.

mozdownload/cli.py
whimboo: That's something I would also do in the argparse PR given that I would have to touch those lines anyway.

mozdownload/cli.py
whimboo: Just to add... reason why those instances are present is that I only copy and pasted the code from one file to the other to minimize code changes for the reviewer. Sorry, that I didn't explicitly mention that.

AutomatedTester:
If you are doing it in a follow up then fine. r+

Issue #318, opened 2015-09-11
Switch from optparse to argparse


whimboo:
We should finally switch from the old optparse module to argparse. I will have a PR up in a bit. For now it will depend on issue #316.

whimboo:
PR #319 got merged.

PR #319, opened 2015-09-11
Switch from optparse to argparse (#318)

whimboo:
This PR will fix issue #318.

### DIFF
--- a/mozdownload/cli.py
+++ b/mozdownload/cli.py
@@ -4,7 +4,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from optparse import OptionParser, OptionGroup
+import argparse
 import os
 import pkg_resources
 import sys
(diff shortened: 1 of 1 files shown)


whimboo:
This PR depends on PR #317. So only the 2nd commit should be obeyed for now.

mozdownload/cli.py
```diff
@@ -0,0 +1,224 @@
...
+                       dest='changeset',
+                       help='Changeset of the try build to download')
+
+    args = parser.parse_args()
+
+    # Gives instructions to user when no arguments were passed
+    if len(sys.argv) == 1:
+        print __doc__
+        parser.format_usage()
+        print 'Specify --help for more information on args. ' \
```
AutomatedTester: unneeded `\`

AutomatedTester:
r+ when comment is addressed

mozdownload/cli.py
whimboo: Well, without using rounded brackets for the print statement the next line would have to start at the same column without indentation. It's kinda hard to read. So I will change the file to use `print('...')` instead.

Issue #320, opened 2015-09-11
Create Factory class for various Scraper instances


whimboo:
Right now it is kinda hard to create the correct scraper. We should make it easier to create a specific one by just having a factory class which accepts various arguments and automatically creates the wanted scraper instance for you.

whimboo:
PR #321 got merged.

PR #321, opened 2015-09-11
Create Factory class for various Scraper instances (#320)

whimboo:
This PR will fix issue #320.

### DIFF
--- a/mozdownload/cli.py
+++ b/mozdownload/cli.py
@@ -9,7 +9,7 @@ import os
 import pkg_resources
 import sys
 
-from . import errors
+from . import factory
 from . import scraper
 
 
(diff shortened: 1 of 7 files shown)


mozdownload/cli.py
```diff
@@ -45,60 +38,41 @@ def cli():
                         default=scraper.BASE_URL,
                         metavar='BASE_URL',
                         help='The base url to be used, default: "%(default)s"')
+    parser.add_argument('--build-number',
+                        dest='build_number',
+                        type=int,
+                        metavar='BUILD_NUMBER',
+                        help='Number of the build (for candidate, daily, and tinderbox builds)')
```
whimboo: Those changes here are only for the correct order, so it applies to the order of arguments for the FactoryScraper.

whimboo:
@davehunt or  @AutomatedTester can one of you please have a look at? Thanks.

Here an example how it could look like in the interactive console: https://pastebin.mozilla.org/8846049

AutomatedTester:
r+

Issue #329, opened 2015-10-21
mozdownload is no longer able to download builds from archive.mozilla.org


whimboo:
Via [bug 1211732](https://bugzilla.mozilla.org/show_bug.cgi?id=1211732) the files on archive.mozilla.org have been moved to S3. That means the HTML is different now and mozdownload fails to download any build.

As a quick workaround we can change the BASE_URL to `http://ftp-origin-scl3.mozilla.org/` to keep mozdownload working. 

I will get this fixed in a bit.

@xabolcs maybe that is what you were seeing?

whimboo:
Maybe [bug 1213721](https://bugzilla.mozilla.org/show_bug.cgi?id=1213721) is a better reference here.

whimboo:
The workaround via PR #330 has been landed. I will releae v 1.18.1 now via issue #331.

A full fix will hopefully be ready later today.

whimboo:
The patch here is easy. The only thing which has been changed is that sub folders are listed with the full URL fragment and not only the final path item:

`/pub/firefox/nightly/2015/10/2015-10-20-04-02-36-ash-l10n/` vs. `2015-10-20-04-02-36-ash-l10n/`.

I have a patch locally which works fine and works for both the old and new link format. I have to check tests before I open a PR. CC'ing @mjzffr, @sydvicious, and @parkouss.

whimboo:
So the parser update is only the first part. I also have to update the scraper because archive.m.o behaves different now and fails if a folder doesn't have a final slash:

http://archive.mozilla.org/pub/firefox/releases/
http://archive.mozilla.org/pub/firefox/releases
http://ftp-origin-scl3.mozilla.org/pub/firefox/releases/

whimboo:
New builds are only uploaded to S3 so we are currently busted with localized builds in Mozmill CI. :( I will work on that issue so we can get it landed ASAP. CC'ing @nthomas-mozilla for info.

whimboo:
Having the need of final slashes for folder now makes it bit complicated for our current mozdownload code base which definitely needs a refactoring. I have all the changes in place now (at least I hope) but testing it is hard given that we only have local tests. As result I will have to write some more tests so we can be sure that all features are working as expected.

whimboo:
PR #335 got merged to master.

whimboo:
[...]

PR #335, opened 2015-10-23
Fix parser and scraper to handle new S3 based archive.mozilla.org (#329)

whimboo:
This PR fixes issue #329. @mjzffr and @sydvicious I need this reviewed. Based on the new tests we don't need manual testing against the remote server anymore.

### DIFF
--- a/mozdownload/parser.py
+++ b/mozdownload/parser.py
@@ -48,7 +48,16 @@ class DirectoryParser(HTMLParser):
 
         for attr in attrs:
             if attr[0] == 'href':
-                self.active_url = attr[1]
+                # Links look like: /pub/firefox/nightly/2015/
+                # We have to trim the fragment down to the last item. Also to ensure we
+                # always get it, we remove a possible final slash first
+                has_final_slash = attr[1][-1] == '/'
+                self.active_url = attr[1].rstrip('/').split('/')[-1]
+
+                # Add back slash in case of sub folders
+                if has_final_slash:
+                    self.active_url = '%s/' % self.active_url
+
                 return
 
     def handle_endtag(self, tag):
--- a/tests/remote/test_fennec.py
+++ b/tests/remote/test_fennec.py
@@ -1,0 +1,93 @@
+#!/usr/bin/env python
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import tempfile
+import unittest
+
+import mozfile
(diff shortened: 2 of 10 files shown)


tests/remote/test_fennec.py
```diff
@@ -0,0 +1,100 @@
...
+    def tearDown(self):
+        mozfile.rmtree(self.temp_dir)
+
+    def test_daily_scraper(self):
+        for test in tests_daily_scraper:
+            scraper = mozdownload.DailyScraper(destination=self.temp_dir,
+                                               log_level='ERROR',
+                                               **test['args'])
+
+            if test.get('url'):
```
mjzffr: Seems like `test['url']` is never actually used. So is this test really testing anything?

whimboo: Good point. I should remove those lines. So the constructor of each scraper is already analyzing the remove server and finds the build. This test is just that we do not see an exception because a path is wrong or a final slash it missing.

whimboo: Updated all new remote test files for that.

mjzffr:
lgtm!

mozdownload/parser.py
```diff
@@ -48,7 +48,16 @@ def handle_starttag(self, tag, attrs):
...
-                self.active_url = attr[1]
+                # Links look like: /pub/firefox/nightly/2015/
+                # We have to trim the fragment down to the last item. Also to ensure we
+                # always get it, we remove a possible final slash first
+                has_final_slash = attr[1][-1] == '/'
+                self.active_url = attr[1].rstrip('/').split('/')[-1]
+
+                # Add back slash in case of sub folders
+                if has_final_slash:
+                    self.active_url = '{}/'.format(self.active_url)
```
sydvicious: This seems a non-obvious way to write:

self.active_url = '%s/' % self.active_url

whimboo: .format() is the new way of doing it. But I decided against that for that file in other lines. So this is just a left-over. I will change it.

sydvicious: Especially since there are a lot of other instances of '%s/' % expression in the rest of the code.

sydvicious:
@whimboo, have you thought instead about checking for and adding a '/' before the connection is established and leaving all of this other code alone?

whimboo:
> @whimboo, have you thought instead about checking for and adding a '/' before the connection is established and leaving all of this other code alone?

This will not work given that at this level you do not know if it is a folder or a file which is referenced. So we have to ensure that the slash gets added at the proper places only.

sydvicious:
@whimboo, r+. LGTM

Issue #328, opened 2015-09-15
Use a shared HTTP/HTTPS session() for retrieving data


whimboo:
As of now we seem to make individual connections to the remote host for querying every tiny information and finally downloading the build. It would be better to share those connections via a session (http://docs.python-requests.org/en/latest/user/advanced/#session-objects). It can result in a more or less good speed improvement.

whimboo:
Downloading an Aurora build shows the following connections being created:

    15:49:38     INFO - Retrieving list of builds from https://archive.mozilla.org/pub/mozilla.org/firefox/nightly/2015/09
    15:49:38     INFO - Starting new HTTPS connection (1): archive.mozilla.org
    15:49:39     INFO - Starting new HTTP connection (1): archive.mozilla.org
    15:49:40     INFO - Starting new HTTPS connection (1): archive.mozilla.org
    15:49:42     INFO - Starting new HTTP connection (1): archive.mozilla.org
    15:49:43     INFO - Found 1 build: 2015-09-14-00-40-20-mozilla-aurora
    15:49:43     INFO - Selected build: 2015-09-14-00-40-20-mozilla-aurora
    15:49:43     INFO - Starting new HTTPS connection (1): archive.mozilla.org
    15:49:45     INFO - Starting new HTTP connection (1): archive.mozilla.org

whimboo:
It will also help us with the authentication given that we only have to set the credentials when the session gets created.

whimboo:
So this can be done by adding the following line to the scraper.py module:

> logging.getLogger("requests").setLevel(logging.DEBUG)

It will enable the maximum amount of logging output. Here how it looks like when we download the latest Nightly build:

> INFO | Retrieving the build status file from https://archive.mozilla.org/pub/mozilla.org/firefox/nightly/latest-mozilla-central/
 INFO | Starting new HTTPS connection (1): archive.mozilla.org
 DEBUG | "GET /pub/mozilla.org/firefox/nightly/latest-mozilla-central/ HTTP/1.1" 200 None
 INFO | Starting new HTTPS connection (1): archive.mozilla.org
 DEBUG | "GET /pub/mozilla.org/firefox/nightly/latest-mozilla-central/firefox-43.0a1.en-US.linux-x86_64.txt HTTP/1.1" 200 99
 INFO | Retrieving list of builds from https://archive.mozilla.org/pub/mozilla.org/firefox/nightly/2015/09
[...]

PR #341, opened 2015-11-10
Use a requests session by scrapper (#328)

parkouss:


whimboo:
So this PR will actually fix issue #328. (explicitly adding the issue # here to set a link to the issue)

mozdownload/parser.py
```diff
@@ -13,8 +13,8 @@
 class DirectoryParser(HTMLParser):
     """Class to parse directory listings"""
 
-    def __init__(self, url, authentication=None, timeout=None):
-        self.authentication = authentication
+    def __init__(self, url, session=None, timeout=None):
+        self.session = session or requests.Session()
```
whimboo: So if we allow `None` for the session it would mean we cannot authenticate anymore and would fail, right?

whimboo:
There is only one thing to clarify. Otherwise this PR looks great! Thanks Julien.

mozdownload/parser.py
parkouss: Indeed, if we pass None (the default value) a session will be automatically created without authentication - still it is of course possible to pass a session that owns authentication data and this is what the scrappers are doing. I wanted to keep the argument optional, in case it was used by some test (yeah, I'm lazy;). If you don't want this behavior, I can force the session to be given (maybe as the first argument).

whimboo: Or we leave in the authentication parameter and pass it into the Session constructor if we have to create a new one?

parkouss: Works for me! I'll do the change that way.

parkouss:
@whimboo updated the PR to address the comment.

mozdownload/parser.py
```diff
@@ -11,10 +11,26 @@
...
+                           None.
+    :param timeout: timeout in seconds used when fetching the directory
+                    content.
+    """
+
+    def __init__(self, url, session=None, authentication=None, timeout=None):
+        if session is None:
+            session = requests.Session()
+            if authentication:
+                session.auth = authentication
```
whimboo: Will the assignment fail if authentication is None? I just wonder about the extra if condition here.

mozdownload/parser.py
```diff
@@ -11,10 +11,26 @@
...
+    :param authentication: a tuple (username, password) to authenticate against
+                           the web server, or None for no authentication. Note
+                           that it will only be used if the given *session* is
+                           None.
+    :param timeout: timeout in seconds used when fetching the directory
+                    content.
+    """
+
+    def __init__(self, url, session=None, authentication=None, timeout=None):
+        if session is None:
```
whimboo: `if not session` should also do it, or?

mozdownload/parser.py
parkouss: I suppose that should work too.

whimboo: Lets try it so that we can simply this code. With the two nits done this PR seems to be complete.

parkouss:
@whimboo updated, should be fine now.

whimboo:
Hm, Travis failed so I retriggered the build.

whimboo:
Travis is green. So I squashed the commits and landed as https://github.com/mozilla/mozdownload/commit/de19d9ed61cb525a66c9ec56d1bfb6b9b2ecf7ee.

Issue #146, opened 2013-08-28
Refactoring of download with retry support


ghost:
There are many places in code which require to download something from remote server.
Moreover, we support case when timeout occurred and then we retry downloading of information.
However, in current state it's really ugly implemented (by many while True) and should be changed into something more readable (contextmanager?) and managable as a pattern.

parkouss:
This is exactly the goal of the redo package I would say: https://pypi.python.org/pypi/redo

I have opened a PR using it, #342.

The output is slightly different, e.g if I raise an exception to try it intentionally:
```
mozdownload --type=daily --branch=mozilla-aurora --platform=win32 --retry-attempts=3
 INFO | Retrieving the build status file from https://archive.mozilla.org/pub/firefox/nightly/latest-mozilla-aurora/
 INFO | Retrieving list of builds from https://archive.mozilla.org/pub/firefox/nightly/2015/11/
 INFO | Found 1 build: 2015-11-08-00-40-59-mozilla-aurora
 INFO | Selected build: 2015-11-08-00-40-59-mozilla-aurora
 INFO | Downloading from: https://archive.mozilla.org/pub/firefox/nightly/2015/11/2015-11-08-00-40-59-mozilla-aurora/firefox-44.0a2.en-US.win32.installer.exe
 INFO | Saving as: /home/jp/dev/mozdownload/2015-11-08-00-40-59-mozilla-aurora-firefox-44.0a2.en-US.win32.installer.exe
 INFO | retry: Calling <function _download at 0x7fad39b54b18> with args: (), kwargs: {}, attempt #2
 INFO | retry: Calling <function _download at 0x7fad39b54b18> with args: (), kwargs: {}, attempt #3
 INFO | retry: Calling <function _download at 0x7fad39b54b18> with args: (), kwargs: {}, attempt #4
 INFO | retry: Giving up on <function _download at 0x7fad39b54b18>
Traceback (most recent call last):
  File "/home/jp/.virtualenvs/mozdownload/bin/mozdownload", line 9, in <module>
    load_entry_point('mozdownload==1.19', 'console_scripts', 'mozdownload')()
  File "/home/jp/dev/mozdownload/mozdownload/cli.py", line 168, in cli
    build.download()
  File "/home/jp/dev/mozdownload/mozdownload/scraper.py", line 313, in download
    errors.TimeoutError))
  File "/home/jp/dev/mozdownload/mozdownload/scraper.py", line 137, in _retry
[...]

PR #342, opened 2015-11-10
 Refactoring of download with retry support (#146)

parkouss:


parkouss:
@whimboo should address #146

mozdownload/scraper.py
```diff
@@ -100,6 +100,7 @@ def __init__(self, destination=None, platform=None,
         logging.basicConfig(format=' %(levelname)s | %(message)s')
         self.logger = logging.getLogger(self.__module__)
         self.logger.setLevel(log_level)
+        logging.getLogger('redo').setLevel(logging.INFO)
```
whimboo: I would use the same log level as for mozdownload. So IMO using `log_level` here would be appropriate. I wonder if we should do the same for requests at the same time.

mozdownload/scraper.py
```diff
@@ -118,28 +119,27 @@ def __init__(self, destination=None, platform=None,
...
-                        message = "Specified build has not been found"
-                        raise errors.NotFoundError(message, e.response.url)
-                    else:
-                        raise
+        self._retry_check_404(self.get_build_info)
+
+    def _retry(self, func, **retry_kwargs):
+        retry_kwargs.setdefault('jitter', 0)
+        retry_kwargs.setdefault('sleeptime', self.retry_delay)
+        retry_kwargs.setdefault('attempts', self.retry_attempts + 1)
```
whimboo: Why do we need one more attempt here?

mozdownload/scraper.py
```diff
@@ -118,28 +119,27 @@ def __init__(self, destination=None, platform=None,
...
+
+    def _retry_check_404(self, func,
+                         err_message="Specified build has not been found",
+                         **retry_kwargs):
+        retry_kwargs.setdefault('retry_exceptions',
+                                (errors.NotFoundError,
+                                 requests.exceptions.RequestException))
+        try:
+            self._retry(func, **retry_kwargs)
+        except requests.exceptions.RequestException as exc:
```
whimboo: So redo already handles `requests.exceptions.RequestException`. Why do we still get it here? I wonder if it is necessary to have this extra wrapper method.

mozdownload/scraper.py
```diff
@@ -326,20 +305,15 @@ def total_seconds(td):
 
                 if log_level <= logging.INFO and content_length:
                     pbar.finish()
-                break
-            except (requests.exceptions.RequestException, errors.TimeoutError), e:
-                if tmp_file and os.path.isfile(tmp_file):
+            except:
+                if os.path.isfile(tmp_file):
```
whimboo: Can we ensure now that tmp_file is always not `None`?

mozdownload/scraper.py
```diff
@@ -326,20 +305,15 @@ def total_seconds(td):
...
-                                     (self.retry_delay))
-                    time.sleep(self.retry_delay)
-                    self.logger.info("Retrying... (attempt %s)" % attempt)
-                if attempt >= self.retry_attempts:
-                    raise
-                time.sleep(self.retry_delay)
+                raise
+
+        self._retry(
+            _download,
```
whimboo: nit: the lines aren't that long yet, so lets move this up right behind the opening brackets.

mozdownload/scraper.py
parkouss: can't be: https://github.com/mozilla/mozdownload/pull/342/files#diff-c9f02456f02d4f3782e1a0f09a60898cL288

whimboo:
Wow with `redo` in place this really saves us a lot of code and makes it way better readable. I really like that. @parkouss I added some comments (which indeed are mostly questions).

mozdownload/scraper.py
parkouss: I believe it was there to have a nicer error formatting. I intended to do the same thing as in https://github.com/mozilla/mozdownload/pull/342/files#diff-c9f02456f02d4f3782e1a0f09a60898cL136

mozdownload/scraper.py
parkouss: because 0 for retry is 0. So that means that if the user pass --retry-count=3, there will be 4 attempts - the first one, then three real retries. This is change in the behavior, but that still make sense to me. We can change that with some hacking if you don't like that change.

mozdownload/scraper.py
parkouss: WFM - be aware that retry is quite verbose;

mozdownload/scraper.py
whimboo: Ah right! Thanks for the pointer.

mozdownload/scraper.py
parkouss: I would not go to do this for requests at the same time though - or maybe we can add it in a separate commit, not squashed at the end ?

mozdownload/scraper.py
whimboo: I see. So we could do that in a follow-up issue without any priority. As it is now it's fully ok. Those are just internals.

mozdownload/scraper.py
whimboo: Better push this out to another issue then.

parkouss:
@whimboo I have fixed the remaining nit.

whimboo:
Looks all fine to me now. Can you please squash the commits and merge to master? Thanks.

parkouss:
Squashed and landed in https://github.com/mozilla/mozdownload/commit/042f08811faea146553b0741124bef4d50741662.

Issue #350, opened 2015-12-09
TestCLICorrectScraper and TestFactoryCorrectScraper fail on OS X


EricRahm:
The best I can tell they have tests for the correct default platform being chosen and we assume that is linux, this assumption is incorrect on OS X.

whimboo:
Fixed with my landed patch for issue #377.

PR #351, opened 2015-12-09
Check for proper platform binary in default tests (#350)

EricRahm:
@whimboo Does this seem reasonable?

### DIFF
--- a/tests/cli/test_correct_scraper.py
+++ b/tests/cli/test_correct_scraper.py
@@ -3,13 +3,23 @@ import os
 import mozfile
 from mozprocess import processhandler
 
+from mozdownload import Scraper
 import mozhttpd_base_test as mhttpd
 
+
+default_latest_binary_names = {
+    'linux': 'firefox-latest-en-US.linux.tar.bz2',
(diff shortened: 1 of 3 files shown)


whimboo:
This will be the fix for issue #350.

tests/cli/test_correct_scraper.py
```diff
@@ -3,13 +3,23 @@
...
+    'mac64': 'firefox-latest.en-US.mac64.dmg',
+    'win32': 'firefox-latest.en-US.win.exe',
+    'win64': 'firefox-latest.en-US.win.exe'
+}
+
 tests = [
     # ReleaseScraper
     {'options': ['-v', 'latest'],
-     'fname': 'firefox-latest.en-US.linux64.tar.bz2'},
+     'fname': default_latest_binary_names[Scraper.detect_platform()]},
```
whimboo: I would suggest that we factor those platform specific tests out into its own test and compare the retrieved filename with the DEFAULT_FILE_EXTENSIONS entry from scraper.py. At the same time we might want to mock the platform so that we can test any of those supported ones.

EricRahm:
Issue #350 indicates this was already fixed, lets go ahead and close this PR.

Issue #355, opened 2016-01-29
Let `latest` for release and candidate point to latest version


whimboo:
With the move to S3 we no longer have symlinks under the latest sub directories. I think we should keep allowing `latest` as version, and have to scrape on our own for the latest version.

whimboo:
PR #356 got merged.

whimboo:
After the merge I noticed that I missed to update the README for the new options. The commit https://github.com/mozilla/mozdownload/commit/3ee65b9565e996f9fc2d200874986006c961551f fixes that.

PR #356, opened 2016-02-01
Enhance --version to accept latest, latest-beta, and latest-esr (#355)

whimboo:
Fix for issue #355.


mozdownload/scraper.py
```diff
@@ -688,15 +714,6 @@ def build_filename(self, binary):
...
                            'EXT': self.extension}
 
-    def download(self):
-        """Download the specified file"""
-
-        try:
-            # Try to download the signed candidate build
-            Scraper.download(self)
-        except errors.NotFoundError, e:
-            self.logger.exception(str(e))
```
whimboo: There are no signed builds around for ages!

whimboo:
@sydvicious or @mjzffr can one of you please review? Thanks.

mozdownload/scraper.py
```diff
@@ -611,19 +618,38 @@ def build_filename(self, binary):
...
+            return parser.entries
+
 
 class ReleaseCandidateScraper(ReleaseScraper):
     """Class to download a release candidate build from the Mozilla server"""
 
-    def __init__(self, version, build_number=None, *args, **kwargs):
-        self.version = version
+    def __init__(self, build_number=None, *args, **kwargs):
         self.build_number = build_number
```
sydvicious: So, the __init__ of this used to take a non-optional version parameter. The ReleaseScraper still does. The other scrapers do not and never did. I may be misunderstanding Python here, but wouldn't the factory creating raise for ReleaseScraper if version was not there, and raise for all of the others if version is there? This seems to be an asymetric API, and since there are no direct calls to these __init__ methods, shouldn't there be a bug here, and shouldn't it have always been there?

mozdownload/scraper.py
```diff
@@ -611,19 +618,38 @@ def build_filename(self, binary):
...
 
-        Scraper.__init__(self, *args, **kwargs)
+        ReleaseScraper.__init__(self, *args, **kwargs)
```
sydvicious: Doesn't this still take a version parameter?

mozdownload/scraper.py
whimboo: Please be aware that the Candidate builds are special release builds. That's why the `ReleaseCandidateScraper` class is based on the `ReleaseScraper` class. Don't ask me why we didn't call its base `__init__()` method, but the one from the `Scraper` class. So that is something I have changed here. Means the version is still mandatory and as you noticed will be picked up by the base class which is `ReleaseScraper`. Regarding other scrapers, none of those need a version but a branch. As result the factory class exists to make it easier to work with mozdownload. It will only pick those parameters which are necessary and known by a class. The others will not be obeyed.

mozdownload/scraper.py
whimboo: The `version` parameter gets forwarded to the base class where it gets picked up.

mozdownload/scraper.py
sydvicious: If this works, I'm OK with it. I just don't really understand how all of this is called, so I can't eyeball it and say it is safe.

sydvicious:
@whimboo r+

mozdownload/scraper.py
whimboo: Yes it does. We have dozen of tests and we can definitely rely on those. Don't worry about the complexity. I know that all this code is not that easy to understand. We clearly need a refactoring at some point soon.

Issue #352, opened 2016-01-19
Make use of tox as test runner for unit tests


whimboo:
We should make use of [tox](https://testrun.org/tox/latest/) as testrunner for our unit tests. It has some nice features like automatic environment creation which would help us a lot here. Maybe we could even integrate coverage reports at the same time.

For an example see the mozharness tox configuration file:
http://mxr.mozilla.org/mozilla-central/source/testing/mozharness/tox.ini

Motwani:
I would like to work on this Issue.

whimboo:
That's great to hear! So I've set you as assignee for this issue. Let me know if you have problems in #ateam or #automation. Thanks!

whimboo:
@Motwani do you already have an idea when you can get started with this issue? I would need an easy way to trigger specific tests only, which is not easily doable right now. Thanks.

Motwani:
I have a fair idea about Tox. I'm not sure how to trigger specific tests but I could make a similar tox file using git instead of mercurial. Which versions of Git should I test with? And how can I trigger specific tests?

whimboo:
The first step for this issue should be add tox and let it create the required virtual environment. We could then also make use of the coverage module similar to mozharness and let the tests being triggered through it.

whimboo:
PR #358 has been merged. As we noticed coverage doesn't really work as of now. So I will file a follow-up issue now.

Aditya, thanks a lot for making this possible! I already love this new feature.

PR #358, opened 2016-02-02
Make use of tox as test runner for unit tests (#352)

Motwani:
Please Review and Suggest Changes @whimboo


whimboo:
To get a reference to the issue... this PR fixes issue #352.

A quick glance over the PR made me happy. It looks great even some thing have to be updated. I will add individual comments beside the appropriate code.

Something I wonder is about the coverage report which doesn't look correct. It only covered the `test.py` file but no actual source under `mozdownload`:

https://travis-ci.org/mozilla/mozdownload/builds/106490318#L251

Did we miss a specific setting?

.gitignore
```diff
@@ -3,3 +3,4 @@
 build
 dist
 tests/venv
+.tox/
```
whimboo: nit: put it at the top so we keep the alphabetical order.

.travis.yml
```diff
@@ -4,11 +4,16 @@ python:
 
 install:
   - pip install pylama
+  - pip install tox
```
whimboo: You can do that in a single line. Just separate `tox` with a space from `pylama`

.travis.yml
```diff
@@ -4,11 +4,16 @@ python:
...
 
 before_script:
   - pylama mozdownload
 
-script: ./run_tests.py
```
whimboo: We don't need this file anymore, so lets get rid of it.

tox.ini
```diff
@@ -0,0 +1,15 @@
...
+envlist = py27
+
+[base]
+deps =
+	coverage
+
+[testenv:py27]
+deps=
+	{[base]deps}
+	-r{toxinidir}/tests/requirements.txt
```
whimboo: I would prefer to get those dependencies added directly here, so that we don't need that extra requirements file anymore.

tox.ini
```diff
@@ -0,0 +1,15 @@
...
+
+commands =
+	coverage run --source mozdownload,test {toxinidir}/tests/test.py
```
whimboo: We don't need the coverage for `test`. It's only `mozdownload`. Also you should be able to get rid of `{toxinidir}` because tox always assumes that as root. So `test/test.py` should be enough.

whimboo:
One more note... please update the settings of your editor so that it doesn't use tabs but 4 spaces instead.

Motwani:
Landed the changes in the PR as stated above. 

I'm not sure about the results generated by coverage. 

Have an ideas @whimboo?

whimboo:
I will have a look myself why coverage is not working as expected.

whimboo:
So I think the problem with coverage here might be that we use the `unittest` package to run our tests. This is code from the standard Python library. As the docs for coverage tell, it does not record anything from those libs. Our modules under `mozdownload` are getting invoked via `unittest`, so coverage maybe checks the stack and ignores us.

I would say that we get rid of coverage for now and leave everything functionality wise as what we had with `run_tests.py`. As a next step following this PR/issue we can investigate if we can make use of another testrunner like nose, pytest, or something else.

So please make the small updates and let me know when it has been done. So that I can merge this PR. Thanks a lot for diving into the tox and travis stuff!

Motwani:
@whimboo have updated the PR.

would love to work on the followup issue ! 

Thanks for this one :)

whimboo:
Thank's a lot. I will get it merged to master right now. For the future it would be good if you would always not squash your commits. Even not for the final version. We usually do that right before the merge.

whimboo:
PR got merged as https://github.com/mozilla/mozdownload/commit/6dd0f66f2b203f5a01d34151d9bbe8f7a2596fce.

Issue #360, opened 2016-02-04
Improve logging and arguments handling


whimboo:
We made a couple of changes in the last months but never really refactored the details about argument handling and logging. Especially the latter is a bit broken (actually since we make use of it) and is blocking me in implementing a good logging strategy on issue #354.

I have local changes working which will improve the situation a lot. I will upload it soon.

whimboo:
PR #361 has been merged.

PR #361, opened 2016-02-04
Improve logging and arguments handling (#360)

whimboo:
This will fix issue #360. Please keep in mind that remote fennec tests are currently broken because API level 11 is no longer supported. I will file a separate issue to get this fixed.

@sydvicious can you please review? Please note that I pushed individual commits so it should make it easier for you to go through the code. Thanks!

### DIFF
--- a/mozdownload/__init__.py
+++ b/mozdownload/__init__.py
@@ -2,26 +2,31 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
+__version__ = '1.19'
+
 from . import cli
 
 from .factory import FactoryScraper
 
-from .scraper import (Scraper,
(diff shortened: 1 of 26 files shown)


mozdownload/__init__.py
```diff
@@ -2,26 +2,31 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
+__version__ = '1.19'
+
```
sydvicious: Is there a reason you don't go ahead and do a version bump here? This is a pretty big change.

whimboo: We never do version bumps while other stuff is still open to be fixed for the next milestone. So the bump will happen with the release issue/PR.

sydvicious:
@whimboo r+

Issue #363, opened 2016-02-04
Fix PEP warnings in source files under mozdownload/


whimboo:
With the latest version of pylama including pep8 I see a huge amount of warnings in my editor for all the source files. This bugs me a lot given that it makes it hard to spot real warnings. So I'm going to get them fixed.

whimboo:
PR #364 has been landed on master.

whimboo:
While the tests were running fine for me locally I noticed that Travis spilled out some other failures. Here mainly because the README.md cannot be read, which indeed is not part of the package. Locally I have made it available via MANIFEST.in but that is indeed not the right way. So a fix for setup.py is necessary. Also moving the definition of __version__ some lines down in __init__.py caused a problem because for the earlier imported cli module it was not defined yet.

So  I have to follow up with another PR to get those issues fixed. Sorry for that.

whimboo:
PR #366 has been merged. So this issue has been finally solved.

Issue #362, opened 2016-02-04
Fennec API level 11 got replaced with API level 15


whimboo:
Since January 29th there are no API level 11 builds produced anymore on mozilla-central. As result the remote Fennec tests are failing. We should get the new API level 15 added and limit API level 11 testing until Jan 28th.

whimboo:
PR #367 has been merged. So this feature will be available with the next 1.20 release.

PR #364, opened 2016-02-04
Fix various PEP warnings (#363)

whimboo:
This PR fixes issue #363. Right now it depends on the PR #361 so more commits are included as necessary. @sydvicious, when reviewing please only check the last commit. Also take into account that the Fennec remote test will still fail. I will work on a fix for issue #362 tomorrow.


whimboo:
I have updated this PR for latest master. So a single commit remains here.

sydvicious:
@whimboo r+

PR #366, opened 2016-02-05
Fix regressions introduced by PEP8 warning fixes (#363)

whimboo:
This PR will fix the failures as introduced with the other PR as landed on master. Sorry for that.

### DIFF
--- a/mozdownload/__init__.py
+++ b/mozdownload/__init__.py
@@ -2,9 +2,10 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from __future__ import absolute_import, print_function, unicode_literals
+from __future__ import absolute_import, unicode_literals
 
 from mozdownload import cli
+from mozdownload.cli import __version__
 from mozdownload.factory import FactoryScraper
 from mozdownload.scraper import (
     Scraper,
@@ -15,8 +16,6 @@ from mozdownload.scraper import (
     TinderboxScraper,
     TryScraper,
 )
-
-__version__ = '1.19'
 
 __all__ = [
     __version__,
--- a/mozdownload/cli.py
+++ b/mozdownload/cli.py
@@ -6,13 +6,15 @@
 
 """Tool to download different Gecko based applications (version {})."""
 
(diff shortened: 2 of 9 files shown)


mozdownload/__init__.py
```diff
@@ -2,7 +2,9 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from __future__ import absolute_import, print_function, unicode_literals
+from __future__ import absolute_import, unicode_literals
```
whimboo: None of our modules are using print anymore. So using `print_function` doesn't make sense.

mozdownload/__init__.py
```diff
@@ -2,7 +2,9 @@
...
+
+__version__ = '1.19'
 
 from mozdownload import cli
```
whimboo: This had to be moved up because the imported `cli` doesn't know about it otherwise.

mozdownload/scraper.py
```diff
@@ -288,7 +288,7 @@ def _download():
                                           maxval=max_value).start()
 
                 with open(tmp_file, 'wb') as f:
-                    for chunk in iter(lambda: r.raw.read(CHUNK_SIZE), ''):
+                    for chunk in r.iter_content(CHUNK_SIZE):
```
whimboo: With `unicode_literals` I got a unicode conversion failure when trying to download the 42.0 release. As given by the requests documentation you should better use its own iterator to prevent the usage of raw strings.

mozdownload/cli.py
```diff
@@ -6,13 +6,15 @@
...
+from __future__ import absolute_import, unicode_literals
 
 import argparse
 import logging
 import os
 
-from mozdownload import __version__, factory, scraper
+from mozdownload import factory, scraper
+
+__version__ = '1.19'
```
whimboo: There is no good place in `__init__.py` for declaring this version without having pylama to blame about it. Given that only `cli.py` is using the variable beside `setup.py`, lets define it here.

whimboo:
@sydvicious can you please review those latest changes? It will fix the regressions as introduced with the other PR for issue #364. Thanks

Issue #368, opened 2016-02-08
Use wptserve for unit tests


whimboo:
Right now the unittests make use of mozhttpd. The development of this HTTP server has been stopped in favor of [wptserve](https://github.com/w3c/wptserve).

whimboo:
PR #369 has been merged.

PR #369, opened 2016-02-09
Use wptserve for unit tests (#368)

whimboo:
This PR will fix issue #368. @mjzffr can you please review it? Thanks

### DIFF
--- a/mozdownload/parser.py
+++ b/mozdownload/parser.py
@@ -69,12 +69,7 @@ class DirectoryParser(HTMLParser):
                 # We have to trim the fragment down to the last item. Also to ensure we
                 # always get it, we remove a possible final slash first
                 url = urllib.unquote(attr[1])
-                has_final_slash = url[-1] == '/'
                 self.active_url = url.rstrip('/').split('/')[-1]
-
-                # Add back slash in case of sub folders
-                if has_final_slash:
-                    self.active_url = '%s/' % self.active_url
 
                 return
 
@@ -85,9 +80,10 @@ class DirectoryParser(HTMLParser):
 
     def handle_data(self, data):
         """Callback when the data of a tag has been collected."""
-        # Only process the data when we are in an active a tag and have an URL
+        # Only process the data when we are in an active a tag and have an URL.
         if not self.active_url:
             return
 
-        if self.active_url == data:
-            self.entries.append(self.active_url.strip('/'))
+        # The visible text can have a final slash so strip it off
+        if data.strip('/') == self.active_url:
+            self.entries.append(self.active_url)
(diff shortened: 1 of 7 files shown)


mozdownload/parser.py
```diff
@@ -89,5 +89,5 @@ def handle_data(self, data):
         if not self.active_url:
             return
 
-        if self.active_url == data:
+        if data in self.active_url:
```
whimboo: We need this change because the actual data can contain a final slash or not, while `active_url` always has it.

tests/directory_parser/test_directory_parser.py
```diff
@@ -20,14 +20,14 @@ def test_init(self):
...
-        parser = DirectoryParser(self.server_address)
+        parser = DirectoryParser(self.wdir)
 
         # relies on the presence of other files in the directory
         # Checks if DirectoryParser lists the server entries
         self.assertNotEqual(parser.entries, [], "parser.entries were not listed")
 
         # path_regex to mozdownload -t release -p win32 -v latest
-        testpath = urljoin(self.wdir, 'directoryparser')
+        testpath = urljoin(self.wdir, 'directoryparser/')
```
whimboo: wptserve will also fail if no final slash has been given for a folder. Maybe @jgraham knows if this is wanted or not.

tests/mozhttpd_base_test.py
```diff
@@ -6,52 +6,41 @@
...
 HERE = os.path.dirname(os.path.abspath(__file__))
 
 
-@mozhttpd.handlers.json_response
-def resource_get(request, objid):
-    """Request handler for JSON responses."""
-    return (200, {'id': objid, 'query': request.query})
-
-
 class MozHttpdBaseTest(unittest.TestCase):
```
whimboo: In the future it might be good to get rid of the HTTP base class but make use of a decorator for those tests which need the server. Performance wise we might also want to start it once and not for each test.

mozdownload/parser.py
mjzffr: Wouldn't it be safer to strip the final slash and then compare? `data in self.active_url` is a very generous condition.

mjzffr:
One question, otherwise looks good.
Ideally, I would use a separate commit for the switch from mozlog to logging.

mozdownload/parser.py
whimboo: Let me check if that would work.

whimboo:
@mjzffr I pushed another commit with some clean-up + a new test for names with spaces. Wanted to be sure that wptserve handles that correctly. Let me know if you like that now.

Regarding the mozlog changes I can redo the commits after I got a r+. Sadly I missed those changes with my recent landing for issue #360. I could re-label it appropriately.

mjzffr:
@whimboo r+

whimboo:
I separated the commits for the wptserve and logging changes. Travis is running now. If results are passing I will get it landed.

Issue #354, opened 2016-01-26
Get daily, tinderbox, and try builds by revision


whimboo:
A requested feature for a long time. Finally with the help of Treeherder we can get this implemented. It's not that easy but with a couple of API requests it can be done. So here an example for one of the latest beta-releases which happened to be:

https://treeherder.mozilla.org/#/jobs?repo=mozilla-beta&revision=5c042dab34a2

Steps to do:

* Get the resultset for the given revision: https://treeherder.mozilla.org/api/project/mozilla-beta/resultset/?revision=5c042dab34a257de85b7eaf2bc81aaad16b8a1b0

* Retrieve the option collection hash to use: https://treeherder.mozilla.org/api/optioncollectionhash/

* With the `results[0]['id']` (2441562) found you can query for all the jobs run for this particular changeset. What we want is to apply filters to reduce the time until we got all results. Keep in mind that we need a hash for Treeherder supported platforms and similar: https://treeherder.mozilla.org/api/project/mozilla-beta/jobs/?result_set_id=2441562&count=2000&exclusion_profile=false&job_type_name=Build&platform=linux32&option_collection_hash=102210fe594ee9b33d82058545b1ed14f4c8206e 

* With the `results[0]['id']` (737977) found you can query for the artifacts of the job: https://treeherder.mozilla.org/api/project/mozilla-beta/artifact/?job_id=737977&name=Job+Info&type=json

* We can use the log URL strip of the filename and take that as base path for all the files contained in this folder.

whimboo:
CC'ing @camd given that we started to talk about a possible path last Friday.

whimboo:
PR #370 has been merged to master.

PR #370, opened 2016-02-09
Get daily, tinderbox, and try builds by revision (#354)

whimboo:
This PR will fix issue #354. It's still missing a unit test for the new treeherder module. I will see if I can do one with a faked treeherder URL by using wptserve for it.

@mjzffr and @sydvicious, if you already want to have a look then feel free.


sydvicious:
@whimboo LGTM r+

whimboo:
To note... I have not split off code into several logical commits. If that helps for reviewing.

mozdownload/treeherder.py
```diff
@@ -2,16 +2,19 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+"""Module to retrieve builds via revision."""
```
whimboo: Ouch! I accidentally landed this file already with https://github.com/mozilla/mozdownload/commit/fc8f22f6b26ed339c375f96fb9f008719f5b3cfa. Not sure why. :(

mozdownload/treeherder.py
```diff
@@ -21,48 +24,77 @@
     'win64': {'build_os': 'win', 'build_architecture': 'x86_64'},
 }
```
whimboo: Something I'm still not clear about is if I should keep this block here or move it into scraper so we have it beside the other constants. It would make it easier for updates, but could confuse because its only relevant for Treeherder. What do you think?

mozdownload/scraper.py
```diff
@@ -944,23 +1012,6 @@ def detect_platform(self):
...
-
-        if not parser.entries:
-            raise errors.NotFoundError('No builds have been found', url)
-
-        self.show_matching_builds(parser.entries)
-
-        self.logger.info('Selected build: %s' % parser.entries[0])
-
-        return (parser.entries, 0)
-
```
whimboo: Via Treeherder we have about 4 months of history for getting try builds. I don't think scraping the server makes sense anymore given that this takes ages to actually get the builds.

whimboo:
This largish PR is ready to get fully reviewed. @mjzffr I would also like to get your feedback. Thanks.

mozdownload/factory.py
```diff
@@ -55,14 +56,14 @@ def __init__(self, scraper_type, **kwargs):
         Tinderbox:
```
mjzffr: debug-build is also for Try?

mozdownload/scraper.py
```diff
@@ -331,17 +333,44 @@ class DailyScraper(Scraper):
...
+
+            # Extract the build folders which are prefixed with the buildid
+            self.builds = [build.rsplit('/', 2)[1] for build in builds]
+            self.show_matching_builds(self.builds)
+
+            # There is only a single build per revision and platform
+            self.build_index = 0
+            self.logger.info('Selected build: %s' % self.builds[self.build_index])
+
+            self.date = datetime.strptime(self.builds[self.build_index][:19],
```
mjzffr: Suggestion, feel free to ignore: it would be helpful to note your assumptions about the data you're working with, like `build`, so it's easier to debug future issues with expressions like `build.rsplit('/', 2)[1]` or `self.builds[self.build_index][:19]`.

mozdownload/factory.py
whimboo: Yes! Good catch. There are also params missing for Tinderbox specific things. I will have a look.

mozdownload/scraper.py
```diff
@@ -721,6 +751,28 @@ def __init__(self, branch='mozilla-central', build_number=None, date=None,
 
     def get_build_info(self):
         """Define additional build information."""
+        # Retrieve build by revision
```
mjzffr: There's a lot in common between `TinderboxScraper.get_build_info` and `DailyScraper.get_build_info`. Want to refactor some of it into `Scraper`?

mozdownload/treeherder.py
mjzffr: `PLATFORM_MAPS` seems to be only used in treeherder.py and I don't know if it makes sense to add a dependency on scraper to treeherder.py. Maybe move all the constants to their own module or to `__init__.py`? In any case, I think it's fine to keep it as is.

mozdownload/treeherder.py
```diff
@@ -21,48 +24,80 @@
...
+        """Retrieve build folders for a given revision with the help of Treeherder.
 
-        # Retrieve the option hash to filter for type of build (opt, and debug for now)
-        option_hash = None
-        for key, values in client.get_option_collection_hash().iteritems():
-            for value in values:
-                if value['name'] == ('debug' if debug_build else 'opt'):
-                    option_hash = key
+        :param revision: Revision of the build to download.
+        :param job_type_name: Name of the job to look for. Defaults to `Build`.
```
mjzffr: What are other valid job type names?

mozdownload/scraper.py
whimboo: Totally agree. I will add a comment here.

mozdownload/scraper.py
whimboo: If we do that the `Scraper` class would also make it available for all other types of scrapers, which is not wanted. We could think about a better inheritance in a follow-up issue. Whereby the big refactoring is still in-front of us. Would that be fine?

mozdownload/treeherder.py
whimboo: Yeah, maybe we can do a `config.py` or such in the future, so we have a central place for all configuration settings.

mozdownload/treeherder.py
whimboo: I will add a list of possible options for now. Will be pushed with the next commit.

mozdownload/scraper.py
mjzffr: Sure

mjzffr:
@whimboo I added a few suggestions; nothing major. Flag me again when/if you want me to take another look.

mjzffr:
looks good

Issue #376, opened 2016-02-11
Add Travis support for OS X


whimboo:
Right now we only run our unit tests on Linux hosts via Travis. We should expand the testing to OS X.

https://docs.travis-ci.com/user/osx-ci-environment/

whimboo:
PR #377 has been merged.

PR #377, opened 2016-02-11
Add Travis support for OS X (#376)

whimboo:
This PR will fix issue #376. It's not ready yet and will need some tweaking.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,35 @@
 language: python
-python:
-  - "2.7"
+
+matrix:
+    include:
+        - python: 2.7
+          os: linux
+          env: TOXENV=py27
+        - language: generic
+          os: osx
+          # 7.2 is OS X 10.11.x
+          # https://docs.travis-ci.com/user/languages/objective-c/#Supported-OS-X-iOS-SDK-versions
+          osx_image: xcode7.2
+          env: TOXENV=py27
 
 install:
-  - pip install pylama tox
+    - if [ "$TRAVIS_OS_NAME" = "osx" ]; then
+        curl -O -s https://bootstrap.pypa.io/get-pip.py;
+        python get-pip.py --user;
+        export PATH=$PATH:~/Library/Python/2.7/bin;
+        pip install --user pylama tox virtualenv;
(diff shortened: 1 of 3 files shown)


tests/cli/test_correct_scraper.py
```diff
@@ -15,8 +15,8 @@
 
 tests = {
     'release': {
-        'args': ['-v', '23.0.1'],
-        'fname': 'firefox-23.0.1.en-US.linux64.tar.bz2',
+        'args': ['-v', '23.0.1', '-p', 'win32'],
```
whimboo: Without the platform specified it would fail on other platforms than linux. To be consistent I used win32 everywhere for now. With mock we can enhance this later.

.travis.yml
```diff
@@ -1,21 +1,34 @@
 language: python
-python:
-  - "2.7"
+
+matrix:
+    include:
+        - python: 2.7
+          env: TOXENV=py27
+        - language: generic
+          os: osx
```
parkouss: For clarity, maybe you can add os: linux above

whimboo:
@parkouss would you mind reviewing this PR? Thanks!

.travis.yml
whimboo: sure

parkouss:
@whimboo r+, this looks good and it seems to work well.

whimboo:
Thanks Julien! I just pushed a commit for the `os: linux` line. Once Travis is happy I will get this PR merged.

tests/cli/test_correct_scraper.py
parkouss: Yes, this looks good - this is just testing a specific os now, but this is the same kind of test, no regression here I think. Yeah you can do a better coverage later (my personal preference is [pytest and its parametrize](http://pytest.org/latest/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions) functionality for this kind of things.

whimboo: Nice I will remember. First we have to switch to pytest which will be done via another issue.

Issue #378, opened 2016-02-11
mozdownload command fails with: "error: unrecognized arguments: mozdownload"


whimboo:
The following change seems to have busted the CLI command of mozdownload:
https://github.com/mozilla/mozdownload/pull/370/files#diff-2204333307f1f8807918caa9eec87871R144

The one cli test we have only checks for the output of the global doc string, which we now always show in the help. That said we have no test coverage for this specific case and so the issue was not recognized. :(

Looks like we need a bustage release.

whimboo:
PR #381 has been merged.

PR #381, opened 2016-02-11
CLI command fails due to broken argv parsing (#378)

whimboo:
This is the bustage fix for issue #378 and also includes a test so that we can be safe it won't happen again.

@mjzffr or @sydvicious could you please review?


mjzffr:
@whimboo r+

sydvicious:
@whimboo r+

Issue #388, opened 2016-02-11
Querying Treeherder by revision fails due to invalid usage of platform value


whimboo:
The unittests haven't caught this problem but a late refactoring on PR #370 broke the filter for the API request of jobs:

"GET /api/project/mozilla-beta/jobs/?platform=build_platform&result_set_id=2441562&job_type_name=Build&option_collection_hash=102210fe594ee9b33d82058545b1ed14f4c8206e&exclusion_profile=False HTTP/1.1" 200 None

Notice the invalid value of `build_platform` as actual platform value.

whimboo:
PR #389 has been merged.

PR #389, opened 2016-02-12
Fix platform filter value for querying Treeherder jobs (#388)

whimboo:
This PR will fix issue #388 and also adds a couple more tests to check that the call chain for Treeherder API endpoints works correctly.


tests/treeherder/test_api.py
```diff
@@ -20,24 +21,52 @@
...
+
+    # Apply query options for the loaded data
+    for option, values in query_options.iteritems():
+        # JSON data uses int values while query options have strings - try to convert
+        try:
+            value = int(values[0])
+        except:
+            value = values[0]
+
+        if api_endpoint == 'jobs':
```
mjzffr: Seems like if `api_endpoint` is not in `('jobs', 'job-log-url') then you can skip this `options,values` loop entirely.

tests/treeherder/test_api.py
```diff
@@ -20,24 +21,52 @@
...
+            data['results'] = [entry for entry in data['results']
```
mjzffr: If I understand correctly, `data['results']` will end up with the entries that do not contradict any of the options, because `None` is ok. In other words, an entry that doesn't contain any of the options would be included. This is surprising to me... when I see a URL with query params I expect all the params to be satisfied in all the results. Am I misunderstanding something?

mozdownload/treeherder.py
```diff
@@ -82,11 +82,11 @@ def query_builds_by_revision(self, revision, job_type_name='Build', debug_build=
 
             # Set filters to speed-up querying jobs
             kwargs = {
-                'platform': self.platform,
                 'option_collection_hash': option_hash,
                 'job_type_name': job_type_name,
                 'exclusion_profile': False,
             }
+            kwargs.update(self.get_treeherder_platform(self.platform))
```
mjzffr: I think it's possible for the result of `self.get_treeherder_platform(self.platform)` to be just `self.platform`, which is a string, right? ...in which case the `update` call would fail, no?

mjzffr:
@whimboo A few things look suspicious to me. Perhaps you can explain them away... :)

tests/treeherder/test_api.py
whimboo: `jog-log-url` has a different data set as `jobs`. For jobs the list to filter is under `results` while for the other it's at the root level.

tests/treeherder/test_api.py
whimboo: Hah, I had the checks at a higher level before without the distinction of `api_endpoint`. Given that we now iterate only over known and available properties we should indeed get rid of it.

whimboo: Oh, no I'm wrong. We cannot given that the loop over options is one level above. So we would still check other query options for other api endpoints here. If we don't want that we would have to do all the option checks inside differnt api_endpoint sections.

mozdownload/treeherder.py
whimboo: That would be a coding error but you are right. We should better check directly and not return the platform. Instead raise a NotSupportedError for it. I will include that in my next commit.

tests/treeherder/test_api.py
mjzffr: I see that but my original comment is about something else: is if the endpoint is neither job-log-url nor jobs then you don't need to loop through options,values at all.

whimboo:
@mjzffr I pushed an update. Please let me know what you think about the loops, and if that is ok for now.

tests/treeherder/test_api.py
whimboo: Oh, now I see. I will check this after dinner.

whimboo:
@mjzffr the latest commit should make you satisfied. :)

mjzffr:
@whimboo r+wc, which I leave completely to your discretion.

Issue #359, opened 2016-02-03
Use pytest or nose as testrunner for unit tests


whimboo:
PR #358 has been shown that coverage doesn't really work. This might be related to the unittest framework we make use of. Maybe we should switch away from that and better use nose or even pytest to run our tests.

https://testrun.org/tox/latest/example/pytest.html

@Motwani, as you told on the other issue you are interested to work on it?

Motwani:
Yes , I would like to work on this issue. 

I am not sure what you meant by "switch away from unittest " ?

whimboo:
My fault given that I also don't know about those details yet. So pytest seem to be a testrunner which supports unittest like tests. So we hopefully shouldn't have to change any of our tests.

whimboo:
PR #394 got merged.

PR #394, opened 2016-03-01
Use pytest as testrunner for unit tests #359

Motwani:
PR fixes issue #359

@whimboo Please Review


tox.ini
```diff
@@ -9,6 +9,10 @@ deps =
     mozprocess >= 0.22
     moztest >= 0.7
     wptserve >= 1.4.0
+    pytest
+    coverage
+    pytest-cov
 
 commands =
-    {toxinidir}/tests/test.py
```
whimboo: Does it mean we can remove test.py now?

tox.ini
```diff
@@ -9,6 +9,10 @@ deps =
     mozprocess >= 0.22
     moztest >= 0.7
     wptserve >= 1.4.0
+    pytest
+    coverage
+    pytest-cov
```
whimboo: Please sort them all alphabetically.

tox.ini
```diff
@@ -9,6 +9,10 @@ deps =
...
 
 commands =
-    {toxinidir}/tests/test.py
+        py.test --cov=mozdownload
+        coverage report
```
whimboo: I would suggest to remove coverage for now given that it still does not work. Also the second line here would result in a duplicated output of the report.

whimboo:
Overall this PR looks great! I tested it locally and I really like the pytest solution. I made some comments which I would like to see fixed before I can merge the PR. Thanks!

tox.ini
Motwani: Yes all 48 tests are collected and run without test.py

tox.ini
```diff
@@ -11,4 +12,4 @@ deps =
     wptserve >= 1.4.0
```
whimboo: manifestparser and moztest are no longer used now given that `test.py` got deleted. So please also kill those dependencies.

Motwani:
@whimboo Updated PR , Please review

tox.ini
```diff
@@ -3,12 +3,11 @@ envlist = py27
 
 [testenv:py27]
 deps =
-    manifestparser >= 1.1
+    pytest
```
whimboo: Put that in alphabetical order and we are fine. Sorry for not noticing that earlier.

tox.ini
```diff
@@ -3,12 +3,11 @@ envlist = py27
...
     mock
     mozfile >= 1.1
     mozprocess >= 0.22
-    moztest >= 0.7
     wptserve >= 1.4.0
 
 commands =
-    {toxinidir}/tests/test.py
+        py.test
```
whimboo: Why those extra blanks? 4 was totally fine.

whimboo:
pytest is still not in the right order but I will fix it before squashing all commits and merging this PR. Thanks Aditya for the work!

whimboo:
While running locally, I also noticed that pytest creates a `.cache` folder in the root of the repository checkout. I will add this to the `.gitignore` file too.

whimboo:
Merged PR to master as https://github.com/mozilla/mozdownload/commit/2be59688fcdebddf22eddc569f6090030b486269.

Issue #395, opened 2016-03-04
Add pylama call to tox config file


whimboo:
Right now we are running pylama only for Travis and AppVeyor. It would be good to add this call to the tox.ini file instead so that we can also benefit from that in local testing.

@Motwani would you like to take this issue?

Motwani:
Yep I would love to work on this :)

whimboo:
This looks great now. I will get it merged in a moment.

whimboo:
Merged the PR as https://github.com/mozilla/mozdownload/commit/0cbc6a183a613ee553a4a1759eae72155f1468b3.

PR #397, opened 2016-03-06
Add pylama call to tox config file (#395)

Motwani:
Solves Issue #395.

@whimboo Please Review.

### DIFF
--- a/pylama.ini
+++ b/pylama.ini
@@ -2,5 +2,8 @@
 format = mccabe,pep8,pep257,pyflakes,pylint
 ignore = C901
 
+[pylama:*/tests]
+skip = 1
+
 [pylama:pep8]
 max_line_length = 100
--- a/pylama.ini
+++ b/pylama.ini
@@ -1,7 +1,7 @@
 [pylama]
 format = mccabe,pep8,pep257,pyflakes,pylint
-skip = tests
 ignore = C901
+ignore = */tests/*
 
 [pylama:pep8]
 max_line_length = 100
--- a/pylama.ini
+++ b/pylama.ini
@@ -1,6 +1,6 @@
 [pylama]
 format = mccabe,pep8,pep257,pyflakes,pylint
-skip = tests/*
+skip = tests
 ignore = C901
 
 [pylama:pep8]
--- a/pylama.ini
+++ b/pylama.ini
@@ -1,7 +1,7 @@
 [pylama]
 format = mccabe,pep8,pep257,pyflakes,pylint
 ignore = C901
-ignore = */tests/*
+skip = */.tox/*,*/tests/*
 
 [pylama:pep8]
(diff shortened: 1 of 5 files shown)


pylama.ini
```diff
@@ -1,5 +1,6 @@
 [pylama]
 format = mccabe,pep8,pep257,pyflakes,pylint
+skip = tests/*
```
whimboo: This doesn't work on Windows given that appveyor failed the build here. I'm sure that `tests` would be totally enough here.

pylama.ini
```diff
@@ -1,5 +1,6 @@
 [pylama]
 format = mccabe,pep8,pep257,pyflakes,pylint
+skip = tests
 ignore = C901
```
whimboo: It looks like that pylama automatically checks for `tox.ini` too, so maybe we can simply combine those two ini files.

whimboo: https://pylama.readthedocs.org/en/latest/#configuration-files

Motwani:
So the latest commit separates pylama command and pytest command as pylama was not ignoring the test folder in WIndows.

Initially it was
py.test --pylama

The commands are
pylama mozdownload
py.test

whimboo:
I would still like to see a single command invocation given that pylama makes itself available as pytest plugin. I will have a look myself.

tox.ini
```diff
@@ -1,13 +1,22 @@
...
+[pylama:pep8]
+max_line_length = 100
+
 [testenv:py27]
 deps =
     mock
     mozfile >= 1.1
     mozprocess >= 0.22
     pytest
+    pylama
```
whimboo: nit: not alphabetically sorted.

tox.ini
```diff
@@ -1,13 +1,22 @@
...
     wptserve >= 1.4.0
 
 commands =
+    pylama mozdownload
     py.test
```
whimboo: So I asked in #pylib for feedback in how to use pylama and the suggestion was to keep it separate. So this will be fine.

tox.ini
```diff
@@ -1,13 +1,22 @@
 [tox]
 envlist = py27
 
+[pylama]
+format = mccabe,pep8,pep257,pyflakes,pylint
+ignore = C901
+
+[pylama:pep8]
+max_line_length = 100
```
whimboo: I would suggest that we move this again and a better place is definitely `setup.cfg`. I figured that out while checking other repositories like https://github.com/davehunt/PyPOM. As you can see there the tox.ini is already complex and we do not want to make it even harder to read.

Motwani:
@whimboo Please Review!

whimboo:
Merged as https://github.com/mozilla/mozdownload/commit/0cbc6a183a613ee553a4a1759eae72155f1468b3

Issue #404, opened 2016-07-25
test android-api-9 needs updateing


Nebelhom:
The current tests throw an error (see below).

This is due to [https://github.com/mozilla/mozdownload/blob/master/tests/remote/test_fennec.py#L17](url)

We are now on api-15, so we need to add a date to this, similar to api-11.

```
======================================================================== FAILURES ========================================================================
__________________________________________________________ FennecRemoteTests.test_daily_scraper __________________________________________________________

self = <test_fennec.FennecRemoteTests testMethod=test_daily_scraper>

    def test_daily_scraper(self):
        for test in tests_daily_scraper:
            mozdownload.DailyScraper(destination=self.temp_dir,
                                     logger=self.logger,
>                                    **test['args'])

tests/remote/test_fennec.py:97: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.tox/py27/local/lib/python2.7/site-packages/mozdownload/scraper.py:344: in __init__
    Scraper.__init__(self, *args, **kwargs)
.tox/py27/local/lib/python2.7/site-packages/mozdownload/scraper.py:133: in __init__
    self._retry_check_404(self.get_build_info)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <mozdownload.scraper.DailyScraper object at 0x7f51a1c51750>
func = <bound method DailyScraper.get_build_info of <mozdownload.scraper.DailyScraper object at 0x7f51a1c51750>>
err_message = 'Specified build has not been found'
retry_kwargs = {'retry_exceptions': (<class 'mozdownload.errors.NotFoundError'>, <class 'requests.exceptions.RequestException'>)}
exc = HTTPError(u'404 Client Error: Not Found for url: https://archive.mozilla.org/pub/mobile/nightly/latest-mozilla-central-android-api-9/',)

[...]

PR #405, opened 2016-07-25
Add date parameter to api-9 test in test_fennec.py (#404)

Nebelhom:
This PR addresses issue #404

### DIFF
--- a/tests/remote/test_fennec.py
+++ b/tests/remote/test_fennec.py
@@ -17,7 +17,8 @@ tests_daily_scraper = [
     # -p android-api-9 --branch=mozilla-central
     {'args': {'application': 'fennec',
               'platform': 'android-api-9',
-              'branch': 'mozilla-central'}},
+              'branch': 'mozilla-central',
+              'date': '2016-01-01'}},
 
     # Support for API level 11 ended on Jan 28th 2016
     # -p android-api-11 --branch=mozilla-central
--- a/tests/remote/test_fennec.py
+++ b/tests/remote/test_fennec.py
@@ -14,6 +14,7 @@ import mozdownload
 
 
 tests_daily_scraper = [
+    # Support for API level 9 ended on Mar 11th 2016
     # -p android-api-9 --branch=mozilla-central
     {'args': {'application': 'fennec',
               'platform': 'android-api-9',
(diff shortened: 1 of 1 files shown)


tests/remote/test_fennec.py
```diff
@@ -17,7 +17,8 @@
     # -p android-api-9 --branch=mozilla-central
     {'args': {'application': 'fennec',
               'platform': 'android-api-9',
-              'branch': 'mozilla-central'}},
+              'branch': 'mozilla-central',
+              'date': '2016-01-01'}},
```
whimboo: Is that the date when the official support ended?

Nebelhom: I did not check for that. I only looked for a valid build... What is the significance of having the last build?

whimboo: For reference and in case we would need that day in the future eg. to determine which service to use for scraping for the build.

Nebelhom: The last one I found was 2016-03-11. I changed the date to that now. Should go through fine now...

Nebelhom:
@whimboo This should do it now. The test ran through fine locally and appveyor gave approval so far.

Please review. Thank you :)

tests/remote/test_fennec.py
```diff
@@ -17,7 +17,8 @@
     # -p android-api-9 --branch=mozilla-central
     {'args': {'application': 'fennec',
               'platform': 'android-api-9',
-              'branch': 'mozilla-central'}},
+              'branch': 'mozilla-central',
+              'date': '2016-03-11'}},
```
whimboo: Looks fine. But please also add a comment as for the API-11 level below.

whimboo:
> NOTE: The other two remote tests (firefox and thunderbird) were skipped. Not sure if that is intentional

Wow, great spot! And no, those two tests should not be skipped. I wonder where this has been set and when!

Nebelhom:
> Wow, great spot! And no, those two tests should not be skipped. I wonder where this has been set and when!

I will open another issue for that...

Nebelhom:
@whimboo and the next round... Thanks for checking once the tests come back positive

whimboo:
Thank you for that patch!

Nebelhom:
> Thank you for that patch!

Oh, we do not squash the pathes into one commit anymore?

whimboo:
Github has changed their ui so it allows it to do directly via the web! That is so cool! :D

PR #411, opened 2016-10-02
Added date to tinderbox filename when human readable date supplied (#181)

### LINKED ISSUE #181: tinderbox builds selected using human-readable date have no unique file name
If I use  "-t tinderbox --date=1383739795", the "1383739795" is preprended instead of the ordinary date.
If I use "-t tinderbox --date=2013-11-06 [-build-num=3]", the file name is not changed, so it is not unique at all.

Nebelhom:
The filename for human-readable code is now prepended by the date format `YYYY-MM-DD-`

If you decide you want a different format or a universal format (e.g. all timestamps) then please let me know.

Closes #181

### DIFF
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -823,16 +823,14 @@ class TinderboxScraper(Scraper):
 
     def build_filename(self, binary):
         """Return the proposed filename with extension for the binary."""
-        if not self.timestamp and self.date:
-            date = self.date.strftime('%Y-%m-%d-')
-        else:
-            date = ''
-        # str(...) necessary to create str type else unicode --> test_tinderbox fail
-        return str('%(TIMESTAMP)s%(BRANCH)s%(DEBUG)s-%(NAME)s' % {
-            'TIMESTAMP': self.timestamp + '-' if self.timestamp else str(date),
+        if hasattr(self, 'builds'):
+            self.timestamp = self.builds[self.build_index]
+
+        return '%(TIMESTAMP)s%(BRANCH)s%(DEBUG)s-%(NAME)s' % {
+            'TIMESTAMP': self.timestamp + '-' if self.timestamp else '',
             'BRANCH': self.branch,
             'DEBUG': '-debug' if self.debug_build else '',
-            'NAME': binary})
+            'NAME': binary}
 
     @property
     def build_list_regex(self):
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -800,7 +800,7 @@ class TinderboxScraper(Scraper):
         if not self.locale_build:
             self.builds, self.build_index = self.get_build_info_for_index(
                 self.build_index)
-            # Assigned to provide unique timestamp in filename for each file
+            # Always force a timestamp prefix in the filename
             self.timestamp = self.builds[self.build_index]
 
     @property
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -743,7 +743,6 @@ class TinderboxScraper(Scraper):
         """Create instance of a tinderbox scraper."""
         self.branch = branch
         self.build_number = build_number
-        self.builds = []
         self.debug_build = debug_build
         self.date = date
         self.revision = revision
@@ -824,7 +823,7 @@ class TinderboxScraper(Scraper):
 
     def build_filename(self, binary):
         """Return the proposed filename with extension for the binary."""
-        if not self.locale_build:
+        if hasattr(self, 'builds'):
             self.timestamp = self.builds[self.build_index]
 
         return '%(TIMESTAMP)s%(BRANCH)s%(DEBUG)s-%(NAME)s' % {
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -800,6 +800,7 @@ class TinderboxScraper(Scraper):
         if not self.locale_build:
             self.builds, self.build_index = self.get_build_info_for_index(
                 self.build_index)
+            # Assigned to provide unique timestamp in filename for each file
             self.timestamp = self.builds[self.build_index]
 
     @property
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -800,6 +800,7 @@ class TinderboxScraper(Scraper):
         if not self.locale_build:
             self.builds, self.build_index = self.get_build_info_for_index(
                 self.build_index)
+            self.timestamp = self.builds[self.build_index]
 
     @property
     def binary_regex(self):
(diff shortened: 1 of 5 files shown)


Nebelhom:
@whimboo Could you please have a look? Thanks.

Do you agree to have two different types of filenames for different date formats?

Thanks a lot :)

mozdownload/scraper.py
```diff
@@ -819,11 +819,16 @@ def binary_regex(self):
 
     def build_filename(self, binary):
         """Return the proposed filename with extension for the binary."""
-        return '%(TIMESTAMP)s%(BRANCH)s%(DEBUG)s-%(NAME)s' % {
-            'TIMESTAMP': self.timestamp + '-' if self.timestamp else '',
+        if not self.timestamp and self.date:
+            date = self.date.strftime('%Y-%m-%d-')
```
whimboo: I think that we should try to always use the timestamp. Shouldn't we be able to here to use it given that we have the subfolder name of the specific build on that day? Please also keep in mind that we can have several tinderbox builds per day, so only the date would cause duplicate filenames.  For the latter we should get another test added.

mozdownload/scraper.py
```diff
@@ -819,11 +819,16 @@ def binary_regex(self):
...
+        else:
+            date = ''
+        # str(...) necessary to create str type else unicode --> test_tinderbox fail
+        return str('%(TIMESTAMP)s%(BRANCH)s%(DEBUG)s-%(NAME)s' % {
```
whimboo: I don't see why we have to use `str()` here. I would suggest that you get rid of it again, and check where the unicode characters are coming from. I believe its from the `self.date.strftime`call which you have added here.

Nebelhom:
@whimboo Hi, I rewrote the code so that now, every filename from tinderbox has a timestamp as part of the filename except for when a specific locale is defined, because as far as I understand it, `self.builds` is not defined then...

Once the tests are completed, could you please review? Thank you :)

mozdownload/scraper.py
```diff
@@ -819,6 +819,9 @@ def binary_regex(self):
 
     def build_filename(self, binary):
         """Return the proposed filename with extension for the binary."""
+        if hasattr(self, 'builds'):
```
whimboo: Maybe we should better initialize the `builds` property to `[]` in the constructor so that it always exist.

whimboo: Localized tinderbox builds will indeed not have a subfolder with a timestamp. Those are all in the same folder, and only the last one is kept. So you better check for `self.locale_build` here.

whimboo:
@Nebelhom this PR might miss the upcoming 1.22 release. But will you be able to finish it off?

whimboo:
@Nebelhom would you mind to check this again?

Nebelhom:
Please ignore this run. Rebased it for further work

Nebelhom:
@whimboo Hi could you please review? To my understanding I took all suggested changes into account.

Thanks :)

mozdownload/scraper.py
```diff
@@ -743,6 +743,7 @@ def __init__(self, branch='mozilla-central', build_number=None, date=None,
         """Create instance of a tinderbox scraper."""
         self.branch = branch
         self.build_number = build_number
+        self.builds = []
```
whimboo: We do not initialize the list for other scrapers in `__init__`. So I don't think we have to do it here too. We could fix it for all in a separate PR, but that's low priority.

mozdownload/scraper.py
```diff
@@ -823,6 +824,9 @@ def binary_regex(self):
 
     def build_filename(self, binary):
         """Return the proposed filename with extension for the binary."""
+        if not self.locale_build:
+            self.timestamp = self.builds[self.build_index]
+
```
whimboo: This would only add the `timestamp` for en-US builds, but localized builds still have the non-prefixed name. See the one test for tinderbox scraper which you didn't update.

So we definitely have to set `self.timestamp` inside of `get_build_info` after the call to `get_build_info_for_index`, which currently only gets executed for `en-US`.

Maybe we shouldn't care about localized builds given that we do not push those to archive.mozilla.org anymore. Those only exist in Taskcluster.

Nebelhom: Hi, could you please let me know whether you want to have this change. The last paragraph stands in strong contrast to the request for change in the paragraphs before...

Sorry, whimboo, it is not clear to me.

mozdownload/scraper.py
Nebelhom: I must have misunderstood you in one of the previous change requests... I thought this is what you asked me to do...
I will remove it again in the next commit...

whimboo: Oh I see. The comment about that I made Oct last year. So please forgive me that I simply forgot about it. Meanwhile I just want to see as much changes as necessary to actually get this feature working without regressing something else. If we want to initialize `self.builds` in `__init__` then we can/should do it for all scrapers at once in a different PR.

mozdownload/scraper.py
whimboo: I think lets leave as it is. We shouldn't care that much about l10n builds on archive.mozilla.org. No-one will be interested in those anyway. So yes, lets leave this code as is.

whimboo:
@Nebelhom, I have seen you pushed some updates lately. Is the PR ready again for a review?

Nebelhom:
@whimboo could you please review after tests are run.

Thanks :)

mozdownload/scraper.py
```diff
@@ -800,6 +800,7 @@ def get_build_info(self):
         if not self.locale_build:
             self.builds, self.build_index = self.get_build_info_for_index(
                 self.build_index)
+            self.timestamp = self.builds[self.build_index]
```
whimboo: If there is no build available, this would raise an exception. Should we allow this? And if yes, how do we handle that?

Nebelhom: My assumption was that https://github.com/Nebelhom/mozdownload/blob/7bdceeacc835d958be17ef802eb21e9f1569b222/mozdownload/scraper.py#L908 would raise an issue if there is no build available, so it would never reach this assignment...

whimboo: Oh, right. So that's fine then. Can you add a comment which explains why we set the timestamp here?

Nebelhom:
@whimboo Thanks for the comments. I have added clarifications in comment form to this PR.

mozdownload/scraper.py
```diff
@@ -800,6 +800,8 @@ def get_build_info(self):
         if not self.locale_build:
             self.builds, self.build_index = self.get_build_info_for_index(
                 self.build_index)
+            # Assigned to provide unique timestamp in filename for each file
```
whimboo: I would rephrase this comment to something like: `Always force a timestamp prefix in the filename`. With that fixed we can get this landed. It looks all fine. Thanks.

Issue #413, opened 2016-10-10
Remote tests broken due to non-existent candidate builds


whimboo:
As noticed in PR #412 the remote tests are broken. The candidate builds for Firefox and Thunderbird which are referenced do no longer exist. It would be good to make it less error-prone by using the latest version but that would require some regex handling for the tests.

I would propose that we simply bump the versions for now and add the latest check in a follow-up.

PR #414, opened 2016-10-10
Fix remote tests due to non-existent candidate builds (#413)

whimboo:
This PR fixes issue #413 by just updating the versions.


whimboo:
Given that there are no code changes and tests are passing now, lets get it landed to unblock other PRs.

Issue #247, opened 2015-02-10
Default branch for Thunderbird should be comm-central and not mozilla-central


whimboo:
Trying to download the latest nightly of Thunderbird fails because mozdownload assume mozilla-central:

> $ mozdownload -t daily -a thunderbird
>   INFO | Retrieving the build status file from https://ftp.mozilla.org/pub/mozilla.org/thunderbird/nightly/latest-mozilla-central/

We might need a new mapping which specified the default branch for each application separately.

son-vyas:
Hi everyone, Can I give it a try?

son-vyas:
Thanks. I will let you know as it is ready

andreheringer:
So, as instructed I'll be taking over this issue.

whimboo:
Thanks a lot! Lets me know if you have questions.

PR #416, opened 2016-10-16
added option to print url, fixes #344

### LINKED ISSUE #344: Add option to print final url to standard output
It would be helpful to have an option to let mozdownload print the final url to stdout. It could be used with pipes to pass into another application or stored to disk.

Prashant-Surya:
Add option to print final url to standard output #344

### DIFF
--- a/mozdownload/cli.py
+++ b/mozdownload/cli.py
@@ -71,6 +71,10 @@ def parse_arguments(argv):
                         choices=scraper.PLATFORM_FRAGMENTS.keys(),
                         metavar='PLATFORM',
                         help='Platform of the application')
+    parser.add_argument('--print-url',
+                        dest='print_url',
+                        action='store_true',
+                        help='Print final URL instead of downloading the file.')
     parser.add_argument('--retry-attempts',
                         dest='retry_attempts',
                         default=0,
@@ -163,7 +167,10 @@ def cli(argv=None):
             scraper_type = 'direct'
 
         build = factory.FactoryScraper(scraper_type, **kwargs)
-        build.download()
+        if kwargs.get('print_url'):
+            logger.info(build.url)
+        else:
+            build.download()
     except KeyboardInterrupt:
         logger.error('Download interrupted by the user')
 
--- a/tests/cli/test_cli_print_url.py
+++ b/tests/cli/test_cli_print_url.py
@@ -1,0 +1,27 @@
+#!/usr/bin/env python
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import subprocess
+import unittest
+
+
+class TestPrintUrlArgument(unittest.TestCase):
+    """Test CLI for print-url argument."""
+
+    def test_print_url_argument(self):
+        url = r'https://raw.github.com/mozilla/mozdownload/master/README.md'
+        try:
+            args = ['mozdownload',
+                '--url={url}'.format(url=url),
+                '--print-url']
+            output = subprocess.check_output(args, stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError as e:
+            output = e.output
+
+        self.assertRegexpMatches(output, url)
+
+if __name__ == '__main__':
+    unittest.main()


mozdownload/cli.py
```diff
@@ -119,6 +119,10 @@ def parse_arguments(argv):
                         help='Version of the application to be downloaded for release '
                              'and candidate builds (special values: %s)' % ', '.join(
                                 scraper.RELEASE_AND_CANDIDATE_LATEST_VERSIONS.keys()))
+    parser.add_argument('--print-url',
+                        action='store_true',
+                        help="Prints final url.")
```
whimboo: I would suggest to use 'Print final URL instead of downloading the file'. Please also add `dest` so we stay in sync with all the other options.

mozdownload/cli.py
```diff
@@ -142,7 +146,6 @@ def parse_arguments(argv):
 def cli(argv=None):
     """CLI entry point for mozdownload."""
     kwargs = parse_arguments(argv or sys.argv[1:])
-
```
whimboo: nit: Unnecessary change. Please revert this empty line.

mozdownload/cli.py
```diff
@@ -163,7 +166,7 @@ def cli(argv=None):
             scraper_type = 'direct'
 
         build = factory.FactoryScraper(scraper_type, **kwargs)
-        build.download()
+        build.download(kwargs['print_url'])
```
whimboo: We should not call download if we only want to print the URL. Add the if condition here instead.

mozdownload/scraper.py
```diff
@@ -266,12 +266,14 @@ def total_seconds(td):
         self.logger.info('Downloading from: %s' % self.url)
         self.logger.info('Saving as: %s' % self.filename)
 
+        if print_url:
+            print self.url
```
whimboo: Please see the line 266 above. We already print the final URL here. The issue itself is that we want to print the URL without downloading the build. Sorry if that was not that clear in the description.

mozdownload/scraper.py
```diff
@@ -266,12 +266,14 @@ def total_seconds(td):
...
+
         tmp_file = self.filename + ".part"
 
         def _download():
             try:
                 start_time = datetime.now()
-
```
whimboo: nit: Same comment as above for the removal of the empty line.

Prashant-Surya:
I'll make those changes and update the pull request.

whimboo:
Sorry that I missed your latest updates for this PR. @Prashant-Surya will you be able to finish it off? I will add some inline comments now.

mozdownload/cli.py
```diff
@@ -119,6 +119,10 @@ def parse_arguments(argv):
                         help='Version of the application to be downloaded for release '
                              'and candidate builds (special values: %s)' % ', '.join(
                                 scraper.RELEASE_AND_CANDIDATE_LATEST_VERSIONS.keys()))
+    parser.add_argument('--print-url',
```
whimboo: nit: All the arguments are sorted alphabetically. Can you please move this block up? Thanks.

Prashant-Surya:
@whimboo sure. I'll make changes

mozdownload/cli.py
```diff
@@ -163,7 +167,10 @@ def cli(argv=None):
             scraper_type = 'direct'
 
         build = factory.FactoryScraper(scraper_type, **kwargs)
-        build.download()
+        if kwargs.get('print_url'):
+            print build.url
```
whimboo: Instead of `print` you should use a call to `logger.info()` so that it get printed correctly.

tests/cli/test_cli_print_url.py
```diff
@@ -0,0 +1,25 @@
...
+    def test_print_url_argument(self):
+        try:
+            args = ['mozdownload',
+                '--url=https://raw.github.com/mozilla/mozdownload/master/README.md',
+                '--print-url']
+            output = subprocess.check_output(args)
+        except subprocess.CalledProcessError as e:
+            output = e.output
+
+        url = r'https://raw.github.com/mozilla/mozdownload/master/README.md'
```
whimboo: Please define `url` above so you do not have to hard-code it in line 17.

whimboo:
Thanks a lot, and sorry again for the delay. If we can get this PR landed soon, it will even make the next release of mozdownload 1.22.

Prashant-Surya:
I'll try to finish it asap

Prashant-Surya:
@whimboo Hi I tried changing the test case. As I'm printing url using logging.info should I use mock for checking URL??

whimboo:
It basically shouldn't matter for the test how the content gets printed. When using `logging.info` it also appears in stdout, and the regex match should find it.

Prashant-Surya:
@whimboo I updated the code can u check?

mozdownload/cli.py
```diff
@@ -146,7 +150,6 @@ def cli(argv=None):
     log_level = kwargs.pop('log_level')
     logging.basicConfig(format='%(levelname)s | %(message)s', level=log_level)
     logger = logging.getLogger(__name__)
-
```
whimboo: nit: Please leave this line untouched. Thanks.

tests/cli/test_cli_print_url.py
```diff
@@ -0,0 +1,28 @@
...
+    def test_print_url_argument(self):
+        try:
+            args = ['mozdownload',
+                '--url=https://raw.github.com/mozilla/mozdownload/master/README.md',
+                '--print-url']
+            output = subprocess.check_output(args, stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError as e:
+            output = e.output
+
+        url = r'https://raw.github.com/mozilla/mozdownload/master/README.md'
```
whimboo: Something which I already mentioned before is that you should define this URL only once. You can pass it into the call to `mozdownload` and also use it for the assertion check later.

Prashant-Surya:
@whimboo Fixed those changes. Sorry. Forgot to add my changes in test file.

tests/cli/test_cli_print_url.py
```diff
@@ -0,0 +1,28 @@
...
+import unittest
+
+
+class TestPrintUrlArgument(unittest.TestCase):
+    """Test CLI for print-url argument."""
+
+    def test_print_url_argument(self):
+        url = r'https://raw.github.com/mozilla/mozdownload/master/README.md'
+        try:
+            url_arg = '--url={url}'.format(url=url)
```
whimboo: Ok, one last thing then we can land it. Please bear with me. :)

There is no need for this extra variable. Just move this line into the list below. Thanks.

whimboo:
Sorry for the late reply but Github didn't inform me about the recent push. @Prashant-Surya best is if you always add an extra comment once you pushed a change, and the PR is ready for the next review.

This patch looks fine now, and I'm going to get it merged. Let me know if you have interests to work on something else, or check our open issues. Thanks!

whimboo:
@Prashant-Surya surely! What do you think about issue #247? It's not that complicated and would be a good next step for you.

Prashant-Surya:
@whimboo But someone already worked on it I think. If it's still is open I'll take it up

whimboo:
He abandoned the work. So you can definitely pick it up.

Prashant-Surya:
Okay. I'll take it up.

Issue #418, opened 2016-11-01
pylama failures


Nebelhom:
In PR #412, we are seeing a bunch of pylama failures in appveyor and travis.

Before we can land the above PR we need to resolve those issues.

Minimal suggestionon what to do are to add at least a config file for pylama including an allowed line-length of 100 chars instead of 80 only.

Nebelhom:
@whimboo This issue should block #412 until resolved.

whimboo:
PR #419 got merged.

PR #412, opened 2016-10-07
Add {posargs} to tox.ini and explained usage in README.md (#398)

Nebelhom:
This PR addresses issue #398 

I encountered the same issue as mentioned. A simple workaround is to not give absolute pathnames but using `tox -- -k <keyword>`. It allows you to sidestep this issue quite easily.

PR #419, opened 2016-11-01
Add pylama config to tox.ini (#418)

Nebelhom:
The PR addresses issue #418 

I only added config data to tox.ini as described in the pylama docs ( https://github.com/klen/pylama#id16 )

I could not fully change it, so let's see if the same error persists in appveyor as before.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,3 +4,9 @@ ignore = C901
 
 [pylama:pep8]
 max_line_length = 100
+
+[pylama:pycodestyle]
+max_line_length = 100
+
+[pylama:pylint]
+max_line_length = 100


Nebelhom:
@whimboo looks like this solved our little problem from #412 

Could you please review?

Thanks

tox.ini
```diff
@@ -1,6 +1,13 @@
 [tox]
 envlist = py27
 
+[pylama:pycodestyle]
+max_line_length = 100
+
+[pylama:pylint]
+max_line_length = 100
+disable = R
+
```
whimboo: Lets not put those configurations into the tox.ini bug use setup.cfg for it. See https://pylama.readthedocs.io/en/latest/#configuration-files.

Also I would prefer to have a general pylama section and not specific ones for each tool used. Does `[pylama]` not work? Also what is `disable = R`? Is that really something we need?

whimboo:
Looks like this PR is blocked on issue #420 now.

setup.cfg
```diff
@@ -4,3 +4,9 @@ ignore = C901
 
 [pylama:pep8]
 max_line_length = 100
+
+[pylama:pycodestyle]
+max_line_length = 100
+
+[pylama:pylint]
+max_line_length = 100
```
whimboo: Cannot we define a single pylama section with global settings? I feel it weird that we have to do it for each individual checker.

Nebelhom: I tried that first, but it would cause pycodestyle failures on line_length

With the way I did it, this did not occur...

Nebelhom:
@whimboo Do you have any idea what is going on with all these errors on tests? Everytime, we try to fix one, another one jumps out...

This connection error I am experiencing on my laptop, too. It cannot come from this PR, since I am only changing some lines in setup.cfg and the error is thrown in `scraper.py`...

whimboo:
Johannes please see my latest comment on this PR. This is basically blocked on issue #420 which we have to get fixed first.

setup.cfg
whimboo: Ok, looks like it is indeed the way to go. It looks fine to me.

whimboo:
Except OS X which is still running, all other platforms have passing tests now. So I would say lets get this PR merged!

Issue #241, opened 2015-01-29
Add support for verifying downloaded file's SHA-512 hash


EricRahm:
Each build has an associated |<product>...<locale><platform>.checksums| file. The SHA-512 hash contained in the file could be used to verify the hash of the download.

Proposal:
  - Add a |--verify| flag
  - Download the checksums file and verify the the SHA-512 hash for the download matches
  - If the hash fails follow the standard retry logic already in place

whimboo:
@EricRahm I assume this hash only applies to the installer binaries to download, but not all the other files including the .zip file for Windows.

EricRahm:
@whimboo There's a hash for every build file. For example: https://ftp.mozilla.org/pub/mozilla.org/firefox/nightly/2015/01/2015-01-29-03-02-02-mozilla-central/firefox-38.0a1.en-US.win64-x86_64.checksums includes both the installer and the zip.

whimboo:
Ah, great. Thanks. Sounds like a good improvement.

parthbakshi:
I would like to work on this bug, whats the best way to work on it.

EricRahm:
@whimboo Do you have any notes for @parthbakshi?

whimboo:
Hi @parthbakshi. Great to hear that you have interests to work on this issue! Sorry for the late response but I was away for a full month.

So the best way is to fork the repository and create a local clone of it. Then you should make yourself familiar with the different kinds of Firefox builds (Nightly, Aurora, Beta, Release, candidate builds, tinderbox builds) which are supported by mozdownload. Once done it would be good to know how this checksum file looks like for those different kinds of builds. After that the download() method of the scraper would have to be extended to being able to also get the checksum file, parse the entries, and do the checks if specified via the command line (--verify).

Let me know if you have further questions. You can also reach me on the moznet IRC network in the #automation channel.

parthbakshi:
Ok i have forked the repo, Is there a doc where i can read about the various builds and how the checksums/SHA-512 actually look?

whimboo:
The different builds are also called `channels`. You can get short introductions on [MDN](https://developer.mozilla.org/en-US/Firefox). Maybe by following the links you will get more comprehensive information. For the checksum you might want to have a look at the [hashlib](https://docs.python.org/2/library/hashlib.html) module documentation of Python.

parthbakshi:
Whenever i try to run mozdownload on my local machine i get
```
[...]

PR #422, opened 2016-12-06
Add support for verifying downloaded file's SHA-512 hash

haouech:
Add support for verifying downloaded file's SHA-512 hash #241

### DIFF
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -800,6 +800,8 @@ class TinderboxScraper(Scraper):
         if not self.locale_build:
             self.builds, self.build_index = self.get_build_info_for_index(
                 self.build_index)
+            # Always force a timestamp prefix in the filename
+            self.timestamp = self.builds[self.build_index]
 
     @property
     def binary_regex(self):
@@ -823,6 +825,7 @@ class TinderboxScraper(Scraper):
 
     def build_filename(self, binary):
         """Return the proposed filename with extension for the binary."""
+
         return '%(TIMESTAMP)s%(BRANCH)s%(DEBUG)s-%(NAME)s' % {
             'TIMESTAMP': self.timestamp + '-' if self.timestamp else '',
             'BRANCH': self.branch,
--- a/tests/cli/test_correct_scraper.py
+++ b/tests/cli/test_correct_scraper.py
@@ -31,7 +31,7 @@ tests = {
 
     'tinderbox': {
         'args': ['-t', 'tinderbox', '-p', 'win32'],
-        'fname': 'mozilla-central-firefox-25.0a1.en-US.win32.installer.exe',
+        'fname': '1374583608-mozilla-central-firefox-25.0a1.en-US.win32.installer.exe',
     },
 
     'try': {
--- a/tests/factory/test_factory_invalid_options.py
+++ b/tests/factory/test_factory_invalid_options.py
@@ -62,7 +62,7 @@ tests = [
     # TinderboxScraper
     {
         'scraper_type': 'tinderbox',
-        'fname': 'mozilla-central-firefox-25.0a1.en-US.win32.installer.exe',
+        'fname': '1374583608-mozilla-central-firefox-25.0a1.en-US.win32.installer.exe',
         'kwargs': {
             'platform': 'win32',
 
--- a/tests/factory/test_factory_scraper.py
+++ b/tests/factory/test_factory_scraper.py
@@ -41,7 +41,7 @@ tests = [
     # TinderboxScraper
     {
         'scraper_type': 'tinderbox',
-        'fname': 'mozilla-central-firefox-25.0a1.en-US.win32.installer.exe',
+        'fname': '1374583608-mozilla-central-firefox-25.0a1.en-US.win32.installer.exe',
         'kwargs': {
             'platform': 'win32',
         },
--- a/tests/tinderbox_scraper/test_tinderbox_scraper.py
+++ b/tests/tinderbox_scraper/test_tinderbox_scraper.py
@@ -17,19 +17,19 @@ import mozhttpd_base_test as mhttpd
 firefox_tests = [
     # -p win32
     {'args': {'platform': 'win32'},
-     'filename': 'mozilla-central-firefox-25.0a1.en-US.win32.installer.exe',
+     'filename': '1374583608-mozilla-central-firefox-25.0a1.en-US.win32.installer.exe',
      'url': 'firefox/tinderbox-builds/mozilla-central-win32/'
             '1374583608/firefox-25.0a1.en-US.win32.installer.exe'},
     # -p win32 --branch=mozilla-central
     {'args': {'branch': 'mozilla-central',
               'platform': 'win32'},
-     'filename': 'mozilla-central-firefox-25.0a1.en-US.win32.installer.exe',
+     'filename': '1374583608-mozilla-central-firefox-25.0a1.en-US.win32.installer.exe',
      'url': 'firefox/tinderbox-builds/mozilla-central-win32/'
             '1374583608/firefox-25.0a1.en-US.win32.installer.exe'},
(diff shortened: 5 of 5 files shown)


whimboo:
@haouech can you please check the test failures and fix those first? Would be good to see passing tests before I dig into the review. If you have questions about how to solve those issues please let me know. Thanks!

whimboo:
@haouech, for information you should run the tests locally before you push any update to the remote repository. See the documentation of mozdownload in how to do that.

haouech:
@whimboo Okey I will run them locally. Thank you

whimboo:
I just want to note that you do not have to close a PR because local changes messed up the branch. You can always force push to your remote branch after resetting the local one. So lets keep this PR open for the remaining work.

whimboo:
Sadly you removed everything from this PR. So I hope you have all the work saved off locally?

haouech:
I had problems merging some commits, so I accidently removed everything. I haven't lost local files although I have backup. I am fixing the test problems now

whimboo:
@haouech is there anything I can help with? Or is it just a matter of time right now to get an updated version uploaded? Thanks.

haouech:
I still get errors after pushing a change, while I every test passes locally. You can check the last commit if you can. I ran tox (v2.2.0) locally and there's no problem

mozdownload/scraper.py
```diff
@@ -242,6 +243,51 @@ def detect_platform(self):
         else:
             return "%s%d" % (mozinfo.os, mozinfo.bits)
 
+    def get_checksum_url(self):
+        """Get the checksum url for current version."""
+        filename = 'SHA512SUMS'
+        url = os.path.split(os.path.split(os.path.split(self.url)[0])[0])[0] + '/' + filename
+        return url
```
whimboo: Hm, where do you see such a checksum file on archive.mozilla.org? I have a hard time finding one.

mozdownload/scraper.py
```diff
@@ -242,6 +243,51 @@ def detect_platform(self):
...
+        return code == 200
+
+    def download_checksum(self):
+        """Download the checksum."""
+        r = self.session.get(self.curl, stream=True)
+        r.raise_for_status()
+
+        with open(self.cfilename, 'wb') as f:
+                for chunk in r.iter_content(CHUNK_SIZE):
+                    f.write(chunk)
```
whimboo: I don't see a reason why we have to write the content of the checksum file to a local file. Just keep it in memory which is also way faster.

mozdownload/scraper.py
```diff
@@ -242,6 +243,51 @@ def detect_platform(self):
...
+
+    def get_checksum_name(self):
+        """Get the name of the platform in the checksum file."""
+        path = (os.path.split(os.path.split(os.path.split(self.url)[0])[0])[1] +
+                '/' + os.path.split(os.path.split(self.url)[0])[1] + '/' +
+                os.path.split(self.url)[1])
+        return path
```
whimboo: Please use the available regex patterns we already have in use for various scraper classes. Basically it should work similarly as when running the following code: `mozdownload -t daily --extension checksums`. So I don't think that we have to reinvent the wheel, and can base on the already existing logic.

mozdownload/scraper.py
```diff
@@ -242,6 +243,51 @@ def detect_platform(self):
...
+            # name = path.split('%20')
+            self.checksum = ''
+            for line in f:
+                content = line.split()
+                checksum = content[0]
+                # self.logger.info('Content[1]: %s' % content[0])
+                if content[1] == name:
+                    # self.logger.info('SHA512 Checksum: %s' % checksum)
+                    self.checksum = checksum
+                    break
```
whimboo: Please use a regex to retrieve the checksum and filename at once from the line.

mozdownload/scraper.py
```diff
@@ -265,6 +311,7 @@ def total_seconds(td):
 
         self.logger.info('Downloading from: %s' % self.url)
         self.logger.info('Saving as: %s' % self.filename)
+        self.logger.info('Directory: %s' % os.path.split(self.url)[0])
```
whimboo: I wouldn't necessarily modify download for the hash. Instead we might have to think about to pass in the `url` to download and do not automatically assume its `self.url`. With that we could re-use it for the real download and the checksum file.

mozdownload/scraper.py
```diff
@@ -555,6 +636,39 @@ def path_regex(self):
...
+        url = urljoin(self.path, filename + '.checksums')
+        return url
+
+    def get_checksum_name(self):
+        """Get the name of the platform in the checksum file."""
+        name = ''
+        if self.platform == 'win64' or self.platform == 'win32':
+            name = 'install/sea/'
+        name = name + self.binary
+        return name
```
whimboo: Similar to what I mentioned before for re-using existing code. Basically it means that we have to replace whatever comes after the platform with the `.checksums` suffix.

mozdownload/scraper.py
```diff
@@ -555,6 +636,39 @@ def path_regex(self):
...
+            self.checksum = ''
+            for line in f:
+                content = line.split()
+                checksum = content[0]
+                # self.logger.info('Content[1]: %s' % content[0])
+                if content[1] == 'sha512' and content[3] == name:
+                    # self.logger.info('SHA512 Checksum: %s' % checksum)
+                    self.checksum = checksum
+                    break
+
```
whimboo: The same code again. Given that all checksum files are identically across build types, please define a single parse method in the basic scraper class.

Bonus point we you can move out checksum related code to its own module, so it will make it easier later to handle checksums for different source locations (eg. archive.mozilla.org, or taskcluster).

whimboo:
As it looks like the failures you mentioned were due to pylama checks in failures in code formatting. As it looks like you were able to get this fixed.

mozdownload/scraper.py
haouech: It is for the candidate build. archive.mozilla.org/pub/firefox/candidates/53.0b4-candidates/build1/ contains the 'SHA512SUMS' file which contains all the checksums

whimboo: Ah, I see. So yes, it's the correct one. But the path or regex of this file has to be a class variable, so that it can be adjusted for each individual scraper class, and the base scraper has the correct one available.

mozdownload/scraper.py
haouech: Do you mean that it's better to add an url parameter to existing download method so we can download both the checksum and the real file with the same method?
I don't download the checksum to a file anymore, I store it in a variable for fast access. This way I think that a separate download method for checksum is simpler.

whimboo: Yes, that's what I meant. Maybe we could also pass the stream to the `download` method, which will use it to store the data inside? This can then be a file output stream, or in case of a checksum an in-memory stream.

whimboo:
@haouech, I just wanted to have a look at your changes, but when I checked the commits there are listed a lot from other PRs which already got landed. I don't expect those to be part of this PR because it makes it very hard to find your own changes.

The reason for that is because you have made all the changes to your local `master` branch, and not a specific implementation branch. I would encourage you to extract your own commits, create a new branch based on latest master, and to import the commits again

Let me know if you have problems.

haouech:
I pushed the wrong branch by mistake so PR got closed. Sorry for that, I pushed the clean changes to master and reopened it. Tell me if you want a new branch that contains my changes or if it is fine to put all on my master.
Thanks

whimboo:
Yes, please create a new branch. Given that your master is up-to-date, you can just branch from its  HEAD. Then switch back to master and remove all of your commits.

haouech:
I've done as you asked, I guess the PR is closed given that master doesn't contain my changes. Is that alright?

haouech:
I opened a new PR [#454](https://github.com/mozilla/mozdownload/pull/454) on the new branch

Issue #428, opened 2017-03-14
Fix binary_regex for candidate/release scraper


whimboo:
As noticed in https://github.com/mozilla/mozmill-ci/issues/828 we have a kinda strict limitation to our regex for the binary of candidate and release builds:

https://github.com/mozilla/mozdownload/blob/master/mozdownload/scraper.py#L601

The problem here is that we are not able to download some of the files when using `--extension` because there is a different file naming convention for Windows and MacOS:

https://archive.mozilla.org/pub/firefox/candidates/52.0b2-candidates/build1/win32/en-US/
https://archive.mozilla.org/pub/firefox/candidates/52.0b2-candidates/build1/mac/en-US/

whimboo:
PR #429 got merged.

PR #429, opened 2017-03-14
Fix binary_regex for candidate/release scraper (#428)

whimboo:
This PR fixes issue #428. @davehunt or @callek, could one of you please have a look?


whimboo:
I just pushed one more commit which adds remote tests for all platforms of release candidate scraper when using --extension. That way we can ensure to not regress something again for mozmill-ci or other tools.

PR #435, opened 2017-04-04
Use a separate branch mapping for applications fixes #247

### LINKED ISSUE #247: Default branch for Thunderbird should be comm-central and not mozilla-central

Prashant-Surya:


mozdownload/scraper.py
```diff
@@ -332,10 +335,16 @@ def show_matching_builds(self, builds):
 class DailyScraper(Scraper):
     """Class to download a daily build from the Mozilla server."""
 
-    def __init__(self, branch='mozilla-central', build_id=None, date=None,
+    def __init__(self, build_id=None, date=None,
```
whimboo: So you have removed this option to the daily scraper? It has to be kept, because other branches should still be able to be passed in. Also we have to update the scraper for tinderbox builds.

Prashant-Surya: I've changed it. Will push in few minutes.

mozdownload/scraper.py
```diff
@@ -332,10 +335,15 @@ def show_matching_builds(self, builds):
 class DailyScraper(Scraper):
     """Class to download a daily build from the Mozilla server."""
 
-    def __init__(self, branch='mozilla-central', build_id=None, date=None,
+    def __init__(self, branch=None, build_id=None, date=None,
                  build_number=None, revision=None, *args, **kwargs):
         """Create an instance of the daily scraper."""
-        self.branch = branch
+        if not branch:
+            application = kwargs.get('application', None)
```
whimboo: The application is already set to `self.application` by `Scraper.__init__()`. So I would suggest that you move the branch setting code after line 352. You can then leave line 338 in-tact, and refer to `self.branch` instead.

mozdownload/scraper.py
```diff
@@ -51,6 +51,9 @@
                            'win32': 'exe',
                            'win64': 'exe'}
 
+DEFAULT_BRANCHES = {'thunderbird': 'comm-central'
```
whimboo: I would suggest you also add firefox and fennec. Just to make it better configurable at some point.

Prashant-Surya: Values should be comm-central for those too?

whimboo: Both have `mozilla-central`, as you hard-coded below. So just move it up here and do not fallback to a default for an unknown version.

Prashant-Surya: So I've to add entries for firefox and fennec in DEFAULT_BRANCHES dictionary right?

whimboo: That's correct. Thanks.

Prashant-Surya: Okay. I'll make those changes.

Prashant-Surya: Should I add b2g also?? I saw it in tests

mozdownload/scraper.py
Prashant-Surya: @whimboo Moving the code below Scraper.__init__() is not working since it is using self.branch which is not set.

mozdownload/scraper.py
whimboo: No, `b2g` is dead, and we should remove it from this repository. I will file a separate issue for this. Thanks for noticing!

mozdownload/scraper.py
whimboo: What do you mean with `below`? What I was referring to is line 352 of the current code. You should not move all the code into `Scraper.__init__()` which certainly will not work, correct.

Prashant-Surya: Not inside. I was talking about moving the code below this  `Scraper.__init__()`.  `Scraper.__init__()` will is using self.branch

whimboo: I don't see that anything in the base scraper class is using the branch. Can you please show the traceback of the failure?

Prashant-Surya:
```self = <mozdownload.scraper.DailyScraper object at 0x7fc7a6119510>
    def get_latest_build_date(self):
        """Return date of latest available nightly build."""
        if self.application not in ('fennec'):
>           url = urljoin(self.base_url, 'nightly', 'latest-%s/' % self.branch)
E           AttributeError: 'DailyScraper' object has no attribute 'branch'
mozdownload/scraper.py:416: AttributeError
```
That was the error when I run tox.

whimboo: Oh, I see. So this is because we do simply too much in the constructor. I think we should move [line 133](https://github.com/mozilla/mozdownload/blob/master/mozdownload/scraper.py#L133) to `self.download()`, where it makes way more sense.

This code change should fix the problem you are seeing above, and should hopefully not break any tests. Can you please verify that?

Prashant-Surya: Sure.

Prashant-Surya: Sorry for the late response. I tried what you've said, it's again resulting in some other error.
```
self = <test_daily_indices.TestDailyScraperIndices testMethod=test_build_indices>

    def test_build_indices(self):
        """Testing for correct build_index in DailyScraper"""
    
        for entry in test_params:
            scraper = DailyScraper(destination=self.temp_dir,
                                   base_url=self.wdir,
                                   logger=self.logger,
                                   **entry['args'])
>           self.assertEqual(scraper.build_index, entry['build_index'])
E           AttributeError: 'DailyScraper' object has no attribute 'build_index'

tests/daily_scraper/test_daily_indices.py:54: AttributeError
```

whimboo: Sorry for being late with my reply too! I somehow completely missed it in my inbox. :(

So the failure occurs because `build_index` will now only be available when you have called download. Maybe in those cases when no `download` is wanted, we should offer a method like `fetch` which would build the appropriate properties, and should be called before accessing any of those.

whimboo:
@Prashant-Surya: I just noticed that we never finished this PR :( Would you still be interested to address the remaining issue?

whimboo:
No feedback anymore. Closing the PR.

PR #444, opened 2017-09-05
Display correct build number for candidate build (#437)

### LINKED ISSUE #437: Wrong build number displayed for candidate build
Just run the following command to see the broken output. Luckily we still download the correct version.

```
% mozdownload -t candidate -v 53.0
INFO | Retrieving list of candidate builds from https://archive.mozilla.org/pub/firefox/candidates/53.0-candidates/
INFO | Found 4 builds: build1, build3, build5, build6
INFO | Selected build: build4
INFO | Downloading from: https://archive.mozilla.org/pub/firefox/candidates/53.0-candidates/build6/mac/en-US/Firefox%2053.0.dmg
INFO | Saving as: /Volumes/data/code/mozdownload/firefox-53.0-build6.en-US.mac.dmg
```

I assume that we print the index of the build but not the actual name what we should do.

ujjwalwahi:
Fixes #437

### DIFF
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -695,8 +695,8 @@ class ReleaseCandidateScraper(ReleaseScraper):
             self.build_index = 0
             self.logger.info('Selected build: build%s' % self.build_number)
         else:
-            self.logger.info('Selected build: build%d' %
-                             (self.build_index + 1))
+            self.logger.info('Selected build: %s' %
+                             (parser.entries[self.build_index]))
 
     @property
     def candidate_build_list_regex(self):


mozdownload/scraper.py
```diff
@@ -695,8 +695,8 @@ def get_build_info(self):
             self.build_index = 0
             self.logger.info('Selected build: build%s' % self.build_number)
         else:
-            self.logger.info('Selected build: build%d' %
-                             (self.build_index + 1))
+            self.logger.info('Selected build: %s' %
```
whimboo: The candidate builds are named "build1", "build2", and so on. We don't want to loose this naming convention here.

ujjwalwahi: Sorry unable to understand the concern. Display format is still same, output now looks like this:
```
INFO | Retrieving list of candidate builds from https://archive.mozilla.org/pub/firefox/candidates/53.0-candidates/
INFO | Found 4 builds: build1, build3, build5, build6
INFO | Selected build: build6
INFO | Downloading from: https://archive.mozilla.org/pub/firefox/candidates/53.0-candidates/build6/linux-x86_64/en-US/firefox-53.0.tar.bz2
```
When build number is not passed from cli argument, we pick the last entry from `parser.entries` list. Items in this list are already in required convention.

Or your concern is to keep the log statement in the following format?
`self.logger.info('Selected build: build%s' % (<build-number>))`

whimboo: Oh, you are right. I missed that the parser entries are already in the correct form. So this works fine for me.

PR #453, opened 2017-10-06
Port progressbar to progressbar2.

nathanielnebel:
Port progressbar to progressbar2, luckily this requires minimal code changes as API compatibility was maintained. I had to open up a new pull request because I noticed I was working in master branch, this has not been rectified

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ def get_version():
                       read('mozdownload', 'cli.py'), re.M)[0]
 
 deps = ['mozinfo >= 0.9',
-        'progressbar == 2.3',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -25,7 +25,7 @@ def get_version():
                       read('mozdownload', 'cli.py'), re.M)[0]
 
 deps = ['mozinfo >= 0.9',
-        'progressbar == 2.3',
+        'progressbar2 >= 3.34.3',
```
whimboo: Good to see that the API we made use of is 100% covered in progressbar2.

Issue #466, opened 2018-03-29
FirefoxRemoteTests.test_tinderbox_scraper fails due to ReadTimeout exception


whimboo:
Currently our tests are failing in `FirefoxRemoteTests.test_tinderbox_scraper` due to:

> ReadTimeout: HTTPSConnectionPool(host='archive.mozilla.org', port=443): Read timed out. (read timeout=60.0)

As it looks like the list of tinderbox builds is too long, and retrieving all the data exceeds the 60s timeout.

We should get this fixed to unblock us from landing new PRs.

whimboo:
Here an example job on Travis:
https://travis-ci.org/mozilla/mozdownload/jobs/348503079

whimboo:
After converting the tests to real pytest format it was easy to discover the failing test. So the problematic test here is the one which retrieves tinderbox builds for mozilla-central on linux64.

I will check more on Tuesday in how we can improve it, or maybe increase the timeout.

whimboo:
The PR in #467 just landed. I will close this issue if it passes tests again.

whimboo:
It passed: https://travis-ci.org/mozilla/mozdownload/builds/361990448

PR #467, opened 2018-03-29
Fix test_tinderbox_scraper for ReadTimeout exceptions

whimboo:
This PR will fix issue #466 but is not ready yet. Lets see how the tests on travis will work.

### DIFF
--- a/tests/remote/test_firefox.py
+++ b/tests/remote/test_firefox.py
@@ -4,318 +4,122 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
-import logging
-import tempfile
+"""Test all scraper classes for Firefox against the remote server"""
+
 import unittest
 import urllib
 
-import mozfile
+import pytest
 
 import mozdownload
 from mozdownload.scraper import BASE_URL
 from mozdownload.utils import urljoin
 
 
-tests_release_scraper = [
-    # -p win32 -v latest
-    {'args': {'application': 'firefox',
-              'platform': 'win32',
-              'version': 'latest'}},
+@pytest.mark.parametrize("args,url", [
+    ({'application': 'firefox', 'platform': 'linux', 'version': 'latest'},
+     None),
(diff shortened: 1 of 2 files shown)


whimboo:
Dave it looks like that the test works way better when using pytest. I think we should get this landed. Does everything look fine?

whimboo:
@mjzffr if you have a minute just to glance over the patch that would be nice. Not sure if @davehunt is around already. The changes are only pytest related so it should be fine for you to review. Thanks.

tests/remote/test_firefox.py
```diff
@@ -4,318 +4,122 @@
...
+    {'branch': 'mozilla-central', 'platform': 'win32', 'locale': 'de'},
+    {'branch': 'mozilla-central', 'platform': 'win32', 'extension': 'txt'},
+    {'branch': 'mozilla-beta', 'platform': 'win32'},
+])
+def test_tinderbox_scraper(tmpdir, args):
+    """Test tinderbox scraper against the remote server."""
+    mozdownload.TinderboxScraper(destination=tmpdir, **args)
+
+
+@unittest.skip('Not testable as long as we cannot grab a latest build')
```
davehunt: No need for unittest, see https://docs.pytest.org/en/latest/skipping.html#skipping-test-functions or should we just delete the test?

tests/remote/test_firefox.py
```diff
@@ -4,318 +4,122 @@
...
+def test_try_scraper():
+    pass
 
-    @unittest.skip('Not testable as long as we cannot grab a latest build')
-    def test_try_scraper(self):
-        pass
 
 if __name__ == '__main__':
     unittest.main()
```
davehunt: We can probably remove this if we're just discovering and running the tests using pytest.

tests/remote/test_firefox.py
whimboo: Thanks. I filed PR #470 to cover both mentioned enhancements.

tests/remote/test_firefox.py
whimboo: Good idea.

PR #483, opened 2018-08-02
Added syntax highlighting to readme

amilajack:


whimboo:
Thank you @amilajack! This looks great, and I will merge it right away. 

I wonder if you could also help us to make our documentation more shiny on PyPI: https://pypi.org/project/mozdownload/. I assume that markdown might be the problem here.

Issue #445, opened 2017-09-05
Invalid build number for candidate scraper should fail


whimboo:
By running the following code, we incorrectly download `build6` instead of failing because there is no `build2` available.
```
% mozdownload -t candidate -v 53.0 --build-number 2
INFO | Retrieving list of candidate builds from https://archive.mozilla.org/pub/firefox/candidates/53.0-candidates/
INFO | Found 4 builds: build1, build3, build5, build6
INFO | Selected build: build6
INFO | Downloading from: https://archive.mozilla.org/pub/firefox/candidates/53.0-candidates/build6/mac/en-US/Firefox%2053.0.dmg
INFO | Saving as: /Volumes/data/code/mozdownload/firefox-53.0-build6.en-US.mac.dmg
```

wardt7:
Hi,

I've started work on this bug and have forked the repository (the bug does still exist after testing it my end); I was just wondering if there's anything I need to know about the ReleaseCandidateScraper itself? I have an idea on how to solve this by checking to see if ('build%s' % self.build_number) returns anything (and if not throwing an exception), but this seems like a super hacky way to solve the issue and I'm sure there's a better way of doing it.

I'm also happy to create a unit test to make sure this doesn't happen again in the future if you wish.

Toby

whimboo:
Thank you for your interest in working on this issue. You are welcome to get started. Here some additional information regarding your questions... You don't have to do an additional request. Instead check the method which prints out which builds were found (see my first comment too). If the wanted build like `build2` in the above case is not present, it has to raise a not found error, and not download the latest build. A unit test will be greatly welcome too. I hope that helps to clarify it.

wardt7:
Hi,

I've made a pull request for this fix. Is it possible for it to be reviewed, as there are a couple of build failures which I think are errors in the tests themselves rather than issues on my end (although I'm not 100% sure)

Thanks,
Toby

whimboo:
Hi @wardt7. Sorry that I missed your comment here. I will immediately have a look at your PR #513.

wardt7:
Hi @whimboo . Apologies but I'm going to have to give up working on this PR as university and family life are massively getting in the way at the moment. More than happy for people to use my fork as a basis for a patch though, and I'll stay subscribed just in case a new assignee has any questions about my fork or the issue itself.

whimboo:
[...]

PR #513, opened 2018-11-10
Invalid build number for candidate scraper should fail #445

wardt7:
Test Plan:
Executed the following commands and checked for desired behavior:
`mozdownload -t candidate -v 53.0 --build-number 2`
`mozdownload -t candidate -v 53.0 --build-number 6`

Note that I'm intending on writing a unit test for this bug to prevent future regressions, however I'm currently having significant issues understanding and executing the test suite. 

Whenever I run the tests using `tox` I get import errors from pretty much all of the tests, saying that they fail to import `mozdownload` - running the test python script directly returns import errors for `mozhttp_base_test`. This is despite all of these files definitely existing, as I can run mozdownload manually just fine.

It also appears that tests/release_candidate_scraper/tests_release_candidate_scraper_indices.py already contains a test for invalid build numbers, but I'm not really sure how it works? The code itself doesn't appear to be checking that an error is raised when an invalid build number is entered so I'm a bit stumped on how to go about that.

Any help would be much appreciated,
Toby

### DIFF
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -696,14 +696,17 @@ class ReleaseCandidateScraper(ReleaseScraper):
         self.show_matching_builds(parser.entries)
         self.builds = parser.entries
         self.build_index = len(parser.entries) - 1
-        if self.build_number and \
-                ('build%s' % self.build_number) in self.builds:
-            self.builds = ['build%s' % self.build_number]
-            self.build_index = 0
-            self.logger.info('Selected build: build%s' % self.build_number)
+        
+        if (self.build_number):
+            if ('build%s' % self.build_number) in self.builds:
+                self.builds = ['build%s' % self.build_number]
+                self.build_index = 0
+                self.logger.info('Selected build: build%s' % self.build_number)
+            else:
+                raise errors.NotFoundError('Selected build does not exist', (u'build%s' % self.build_number))
         else:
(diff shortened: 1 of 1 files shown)


wardt7:
Currently there's a fair few failures when testing the build. However, I think these are issues with the unit tests themselves from what I've seen so far:

- `test_release_candidate_scraper_indices.py` is probably broken right now as it turns out there is an intentional invalid build check in there in the first place. However, the way it is evaluating that is almost certainly broken now as it isn't checking to see if it gets a NotFoundError. Whilst I was going to do the unit test in a different pull, I'm happy to try and modify this test instead to get it working.

- The failures in `tests/remotes/devedition.py` are probably due to a mismatch in build number. The URLs for the files seem to use build 3, which is different to the build number supplied (1). It would therefore make sense that the test fails as build 1 wouldn't exist at that URL. I can't confirm this though until I get my test suite properly running, and it is just a guess as I'm not familiar with the devedition build/test.

whimboo:
Ok, lets get to it step by step. Were you meanwhile able to execute `tox` locally, and run the full test suite? If not please attach a log of the tox output. I would like to have a look at it.

Once tests are running, they should all pass without any modification of the code. If that is not the case we would have to check that first.

whimboo:
As @wardt7 noted on the referenced issue there is no time for him/her to get the PR finished. As such we will have to close it. It's sad because it was close to be finished.

Issue #468, opened 2018-04-04
Convert unit tests to plain pytest


whimboo:
All of our tests for mozdownload still make use of the Python unittest framework, which is not necessary because pytest is used to execute all the tests. As such we should get rid of the test classes, and instead use fixtures. In https://github.com/mozilla/mozdownload/pull/467/commits/e0c9484b3a875cde305c59bd7ed5789c5410d179 I already converted `remote/test_firefox.py`.

We should do the same for all the remaining tests in the following sub folders: 

- [x] base_scraper
- [x] cli
- [x] daily_scraper
- [x] direct_scraper
- [x] directory_parser
- [x] factory
- [x] release_candidate_scraper
- [x] release_scraper
- [x] remote
- [x] tinderbox_scraper
- [x] treeherder
- [x] try_scraper

If you want to work on the fennec or thunderbird test please let us know here, and push a PR for only one of both at a time. That way we can divide the work. Thanks.

Once all of the above are done we can safely remove the `MozHttpdBaseTest` class in https://github.com/mozilla/mozdownload/blob/master/tests/mozhttpd_base_test.py.

danishprakash:
Hi @whimboo, as is visible in this issue that there has been a lot of activity on this, is this issue still active, can I take this up?

Utkarsh1308:
Can I work on this issue?

Utkarsh1308:
Please review https://github.com/mozilla/mozdownload/pull/521

Utkarsh1308:
What exactly do I have to do? Sorry I'm fairly new to open source. Do you mean converting other tests to pytests now that we are done with the remote tests?

whimboo:
That is correct. If you don't mind you could pick another subfolder, and convert all of the tests to pytest. If you face problems please let me know. Thanks!

Utkarsh1308:
Thanks! Then I would be starting with release_candidate_scraper tests

Utkarsh1308:
Could you help me with this error - NotFoundError: Specified build has not been found: https://archive.mozilla.org/pub/firefox/candidates/23.0.1-candidates/

Here are the changes I made:
https://github.com/Utkarsh1308/mozdownload/blob/bugFix/tests/release_candidate_scraper/test_release_candidate_scraper_indices.py

whimboo:
As the failure shows you are using a remote HTTP server for the tests now which doesn't work. Everything beside the remote tests are using a local server instead. You removed it with your latest changes. 

[...]

PR #521, opened 2018-12-22
test_thunderbird.py: Convert unit test to plain pytest #468

Utkarsh1308:


whimboo:
Hi @Utkarsh1308! Sorry for the late reply but I was away during the Christmas holidays. Now I have seen your PR and it looks fantastic! There is nothing which needs to be further changed, given that also the tests are passing. 

Hurray for having all remote tests driven by pure pytest now! Thanks a lot!

PR #523, opened 2019-01-09
Change test_release_candidate_scraper_indices.py to pytest #468

Utkarsh1308:


tests/conftest.py
```diff
@@ -0,0 +1,16 @@
...
+
+WDIR = 'data'
+HERE = os.path.dirname(os.path.abspath(__file__))
+
+@pytest.fixture
+def setUp():
+    httpd = server.WebTestHttpd(port=8080,
+                                        doc_root=os.path.join(HERE, WDIR),
+                                        host='127.0.0.1')
+    wdir = 'http://{}:{}'.format(httpd.host, httpd.port)
```
whimboo: Instead of continuing in using our custom `wdir` variable I would suggest we switch to make use of the public method [get_url()](https://dxr.mozilla.org/mozilla-central/rev/c2593a3058afdfeaac5c990e18794ee8257afe99/testing/web-platform/tests/tools/wptserve/wptserve/server.py#709-715) inside the tests.

That way we also save us from not having the fixture to return an additional value.

tests/release_candidate_scraper/test_release_candidate_scraper_indices.py
```diff
@@ -6,53 +6,22 @@
...
 
-class ReleaseCandidateScraperTest_build_indices(mhttpd.MozHttpdBaseTest):
-    """test indices in mozdownload scraper class"""
-
-    def test_build_indices(self):
-        """Testing indices in choosing builds for ReleaseCandidateScraper"""
-
-        for entry in test_params:
-            scraper = ReleaseCandidateScraper(destination=self.temp_dir,
-                                              base_url=self.wdir,
```
whimboo: Note that the base url has to be specified to the server. Do that with the before-mentioned call to `httpd.get_url()`.

whimboo:
One note before reviewing the PR again. Please do not force push your changes in the future. this will make it very hard to see the differences compared to the last revision. Once we are ready to land the PR all commits will automatically be squashed. Thanks

tests/conftest.py
```diff
@@ -0,0 +1,17 @@
+import pytest
+
+import os
```
whimboo: `os` is a global module and has to be listed first. You may have mixed it up with pytest.

tests/conftest.py
```diff
@@ -0,0 +1,16 @@
...
+import pytest
+
+WDIR = 'data'
+HERE = os.path.dirname(os.path.abspath(__file__))
+
+@pytest.fixture
+def setUp():
+    httpd = server.WebTestHttpd(port=8080,
+                                        doc_root=os.path.join(HERE, WDIR),
+                                        host='127.0.0.1')
```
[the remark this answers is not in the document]
whimboo: Please note that when you fix a review comment you might also want to mark it as resolved on the PR. I did that for you now.

tests/conftest.py
whimboo: You didn't address that comment correctly. We don't want a global `wdir` variable.

tests/release_candidate_scraper/test_release_candidate_scraper_indices.py
```diff
@@ -5,54 +5,24 @@
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
 from mozdownload import ReleaseCandidateScraper
+import conftest
 
-import mozhttpd_base_test as mhttpd
+import pytest
```
whimboo: The import of pytest has to be before any mozdownload import.

tests/release_candidate_scraper/test_release_candidate_scraper_indices.py
whimboo: You didn't address that comment.

tests/conftest.py
Utkarsh1308: In my tests. How am I supposed to use get_url() on httpd as it is a function. Do I need to make a httpd variable in my tests as well?

whimboo: Oh, I missed that you wrongly return the return value of `httpd.start()`. Instead you have to return `httpd`. That way you have access to the method.

tests/release_candidate_scraper/test_release_candidate_scraper_indices.py
Utkarsh1308: Please review the changes I've made. I think I've resolved this.

tests/conftest.py
```diff
@@ -0,0 +1,13 @@
+import os
+import pytest
```
whimboo: nit: please add an empty line between both to separate system and global imports.

tests/conftest.py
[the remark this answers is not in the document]
whimboo: Please fix the indentation here. Also please move the `host` argument before the port one. It should look like this:

```
    httpd = server.WebTestHttpd(
        host="127.0.0.1",
        port=8080,
        doc_root=os.path.join(HERE, 'data'),
        )
```

tests/conftest.py
```diff
@@ -0,0 +1,13 @@
...
+from wptserve import server
+
+
+@pytest.fixture
+def httpd():
+    HERE = os.path.dirname(os.path.abspath(__file__))
+    httpd = server.WebTestHttpd(port=8080,
+                                        doc_root=os.path.join(HERE, 'data'),
+                                        host='127.0.0.1')
+    return httpd
```
whimboo: Why did you remove `httpd.start()` and the yield? Both are still needed to first make the server available to the test, and to clean it up afterward.

tests/release_candidate_scraper/test_release_candidate_scraper_indices.py
```diff
@@ -3,56 +3,32 @@
...
+    ({'application': 'firefox', 'build_number': '3', 'platform': 'mac', 'version': '23.0.1'},
+     '0', '3', ['build3']),
+    ({'application': 'firefox', 'build_number': '2', 'platform': 'linux', 'version': '23.0.1'},
+     '1', '2', ['build1', 'build3']),
+])
+
+def test_build_indices(httpd, tmpdir, args, build_index, build_number, builds):
+    """Testing indices in choosing builds for ReleaseCandidateScraper"""
+
+    return httpd.start(block=False)
```
whimboo: This would return from the the tests immediately without running them. Also see my above comment.

tests/release_candidate_scraper/test_release_candidate_scraper_indices.py
```diff
@@ -3,56 +3,32 @@
...
-
-        for entry in test_params:
-            scraper = ReleaseCandidateScraper(destination=self.temp_dir,
-                                              base_url=self.wdir,
-                                              logger=self.logger,
-                                              **entry['args'])
-            self.assertEqual(scraper.build_index, entry['build_index'])
-            self.assertEqual(scraper.build_number, entry['build_number'])
-            self.assertEqual(scraper.builds, entry['builds'])
+    httpd.stop()
```
whimboo: This is the task of the fixture, and should not be done by each and every test.

tests/conftest.py
Utkarsh1308: I won't be able to return an argument in the generator if there is a yield. How am I supposed to return httpd along with a yield httpd.start()?

whimboo: No, you have to yield `httpd` and not the return value from `httpd.start()` as I already mentioned in an earlier review comment. So it will look like the following:

```
httpd.start(block=False)
yield httpd
httpd.stop()
```

That will start the server, use `httpd` for the test, and once that has been finished stops the server.

Utkarsh1308: Please review the changes

tests/release_candidate_scraper/test_release_candidate_scraper_indices.py
```diff
@@ -3,56 +3,28 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
+import pytest
+
+from wptserve import server
```
whimboo: This import isn't used and need to be removed.

tests/release_candidate_scraper/test_release_candidate_scraper_indices.py
```diff
@@ -3,56 +3,28 @@
...
 
 from mozdownload import ReleaseCandidateScraper
+import conftest
```
whimboo: You don't have to import conftest. This is magically done by pytest.

tests/release_candidate_scraper/test_release_candidate_scraper_indices.py
whimboo: Yes, this looks fine.

Utkarsh1308:
Please review the changes

whimboo:
Great work! As it looks like you learned quite a bit about pytest and fixtures on this PR. I hope you are still interested in picking some more? Maybe all the remaining tests from that release candidate folder?

PR #528, opened 2019-01-22
TestScraper: port unittest to pytest #468

rgpt:
- Porting UnitTest code of tests/Pyscraper Folder using pytest library
-  Linting done on both the files.

### DIFF
--- a/tests/try_scraper/test_invalid_revision.py
+++ b/tests/try_scraper/test_invalid_revision.py
@@ -5,24 +5,18 @@
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
 from mock import patch
-
+import pytest
 from mozdownload import TryScraper
 import mozdownload.errors as errors
-import mozhttpd_base_test as mhttpd
 
 
-class TestTryScraperInvalidParameters(mhttpd.MozHttpdBaseTest):
-    """test mozdownload TryScraper class with invalid parameters"""
+@patch('mozdownload.treeherder.Treeherder.query_builds_by_revision')
+def test_scraper(query_builds_by_revision, httpd, tmpdir):
+    """Testing download scenarios with invalid parameters for TryScraper"""
+    query_builds_by_revision.return_value = []
 
-    @patch('mozdownload.treeherder.Treeherder.query_builds_by_revision')
-    def test_scraper(self, query_builds_by_revision):
-        """Testing download scenarios with invalid parameters for TryScraper"""
-        query_builds_by_revision.return_value = []
-
-        with self.assertRaises(errors.NotFoundError):
-            TryScraper(destination=self.temp_dir,
-                       base_url=self.wdir,
-                       logger=self.logger,
-                       platform='win32',
-                       revision='abc'
-                       )
+    with pytest.raises(errors.NotFoundError):
+        assert TryScraper(destination=tmpdir,
+                          base_url=httpd.get_url(),
+                          platform='win32',
+                          revision='abc')
--- a/tests/try_scraper/test_try_scraper.py
+++ b/tests/try_scraper/test_try_scraper.py
@@ -6,73 +6,55 @@
 
 import os
 import urllib
-
+import pytest
 from mock import patch
-
 from mozdownload import TryScraper
 from mozdownload.utils import urljoin
-import mozhttpd_base_test as mhttpd
 
-firefox_tests = [
+@pytest.mark.parametrize("args,filename,url", [
     # -p mac64 --revision=8fcac92cfcad
-    {'args': {'platform': 'mac64',
-              'revision': '8fcac92cfcad'},
-     'filename': '8fcac92cfcad-firefox-38.0a1.en-US.mac.dmg',
-     'url': 'firefox/try-builds/test-user@mozilla.com-8fcac92cfcad/'
-            'try-macosx64/firefox-38.0a1.en-US.mac.dmg'},
+    ({'platform': 'mac64', 'revision': '8fcac92cfcad'},
+     '8fcac92cfcad-firefox-38.0a1.en-US.mac.dmg',
+     'firefox/try-builds/test-user@mozilla.com-8fcac92cfcad/'
+     'try-macosx64/firefox-38.0a1.en-US.mac.dmg'),
     # -p mac --revision=8fcac92cfcad
-    {'args': {'platform': 'mac',
-              'revision': '8fcac92cfcad'},
-     'filename': '8fcac92cfcad-firefox-38.0a1.en-US.mac.dmg',
-     'url': 'firefox/try-builds/test-user@mozilla.com-8fcac92cfcad/'
-            'try-macosx64/firefox-38.0a1.en-US.mac.dmg'},
+    ({'platform': 'mac', 'revision': '8fcac92cfcad'},
+     '8fcac92cfcad-firefox-38.0a1.en-US.mac.dmg',
+     'firefox/try-builds/test-user@mozilla.com-8fcac92cfcad/'
+     'try-macosx64/firefox-38.0a1.en-US.mac.dmg'),
     # -a firefox -p linux64 --revision=8fcac92cfcad
-    {'args': {'platform': 'linux64',
-              'revision': '8fcac92cfcad'},
-     'filename': '8fcac92cfcad-firefox-38.0a1.en-US.linux-x86_64.tar.bz2',
-     'url': 'firefox/try-builds/test-user@mozilla.com-8fcac92cfcad/'
-            'try-linux64/firefox-38.0a1.en-US.linux-x86_64.tar.bz2'},
+    ({'platform': 'linux64', 'revision': '8fcac92cfcad'},
+     '8fcac92cfcad-firefox-38.0a1.en-US.linux-x86_64.tar.bz2',
+     'firefox/try-builds/test-user@mozilla.com-8fcac92cfcad/'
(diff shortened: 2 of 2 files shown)


tests/try_scraper/test_invalid_revision.py
```diff
@@ -5,24 +5,18 @@
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
 from mock import patch
```
whimboo: Would you mind replacing the usage of the plain mock package with https://pypi.org/project/pytest-mock/? This would also apply to the other files.

tests/try_scraper/test_invalid_revision.py
```diff
@@ -5,24 +5,18 @@
...
 
-class TestTryScraperInvalidParameters(mhttpd.MozHttpdBaseTest):
-    """test mozdownload TryScraper class with invalid parameters"""
 
-    @patch('mozdownload.treeherder.Treeherder.query_builds_by_revision')
-    def test_scraper(self, query_builds_by_revision):
-        """Testing download scenarios with invalid parameters for TryScraper"""
-        query_builds_by_revision.return_value = []
+@patch('mozdownload.treeherder.Treeherder.query_builds_by_revision')
+def test_scraper(query_builds_by_revision, httpd, tmpdir):
```
whimboo: Given that we loose the class name, this test needs a clear name now. Also lets add the patch argument as last one so that fixtures always come first.

tests/try_scraper/test_invalid_revision.py
```diff
@@ -5,24 +5,18 @@
...
 
-        with self.assertRaises(errors.NotFoundError):
-            TryScraper(destination=self.temp_dir,
-                       base_url=self.wdir,
-                       logger=self.logger,
-                       platform='win32',
-                       revision='abc'
-                       )
+    with pytest.raises(errors.NotFoundError):
+        assert TryScraper(destination=tmpdir,
```
whimboo: What are you trying to assert here?

tests/try_scraper/test_try_scraper.py
```diff
@@ -6,73 +6,55 @@
...
+import pytest
 from mock import patch
-
 from mozdownload import TryScraper
 from mozdownload.utils import urljoin
-import mozhttpd_base_test as mhttpd
 
-firefox_tests = [
+@pytest.mark.parametrize("args,filename,url", [
     # -p mac64 --revision=8fcac92cfcad
```
whimboo: nit: we can get rid of those comments. So just remove them.

tests/try_scraper/test_try_scraper.py
```diff
@@ -6,73 +6,55 @@
...
-    {'args': {'platform': 'mac64',
-              'revision': '8fcac92cfcad'},
-     'filename': '8fcac92cfcad-firefox-38.0a1.en-US.mac.dmg',
-     'url': 'firefox/try-builds/test-user@mozilla.com-8fcac92cfcad/'
-            'try-macosx64/firefox-38.0a1.en-US.mac.dmg'},
+    ({'platform': 'mac64', 'revision': '8fcac92cfcad'},
+     '8fcac92cfcad-firefox-38.0a1.en-US.mac.dmg',
+     'firefox/try-builds/test-user@mozilla.com-8fcac92cfcad/'
+     'try-macosx64/firefox-38.0a1.en-US.mac.dmg'),
```
whimboo: Would the `url` fit into a single line now by not reaching the 100 char limit? We don't have that much indentation anymore, and would be helpful for readability.

tests/try_scraper/test_try_scraper.py
```diff
@@ -6,73 +6,55 @@
...
-            self.assertEqual(scraper.filename, expected_filename)
-            self.assertEqual(urllib.unquote(scraper.url),
-                             urljoin(self.wdir, entry['url']))
+    ({'platform': 'win64', 'revision': '8fcac92cfcad'},
+     '8fcac92cfcad-firefox-38.0a1.en-US.win64.installer.exe',
+     'firefox/try-builds/test-user@mozilla.com-8fcac92cfcad/'
+     'try-win64/firefox-38.0a1.en-US.win64.installer.exe')
+])
+
+
```
whimboo: Please remove those extra lines so that decorators are directly connected to the function definition.

tests/try_scraper/test_try_scraper.py
```diff
@@ -6,73 +6,55 @@
...
+@patch('mozdownload.treeherder.Treeherder.query_builds_by_revision')
+def test_scraper(mock_query_builds_by_revision, httpd, tmpdir, args, filename, url):
+    """Testing various download scenarios for TryScraper"""
+    mock_query_builds_by_revision.return_value = [
+        '/firefox/try-builds/test-user@mozilla.com-8fcac92cfcad/try-foobar/'
+    ]
+    baseurl = httpd.get_url()
```
whimboo: The getter for the url is cheap, so simply get rid of this extra variable declaration/usage.

tests/try_scraper/test_invalid_revision.py
rgpt: I also wanted to use it. But was not sure whether to use it as mock was used in other files too. Your comment helps on that matter.

tests/try_scraper/test_invalid_revision.py
rgpt: This verifies whether using the provided parameters TryScraper can be instantiated or not. Try Scrapper internally calls Scraper.__init__ which checks the provided parameters. So for example in `platform` if i provide `win32random` mistakenly then the test will fail.

rgpt:
@whimboo updated PR. Please review.

tests/try_scraper/test_invalid_revision.py
whimboo: Sure, but I don't see the need for an assert here. If it fails `__init__` will raise an exception, which then bubbles up to the test itself, because `pytest.raises` only checks for `NotFoundError`. The assert won't be run at any time.

requirements/tests.txt
```diff
@@ -5,3 +5,4 @@ mozprocess >= 0.22
 pytest==4.1.0
 pytest-cov==2.6.1
 wptserve >= 1.4.0
+pytest-mock >= 1.10.0
```
whimboo: Please keep the alphabetical order of inputs and make it strictly requiring this specific version.

rgpt: Should it be done for wptserve too?. Or is there any specific reason to keep it as >=.

tests/try_scraper/test_invalid_revision.py
rgpt: I agree, I will remove `assert` from here. Moreover, I was also wondering about the type of arguments we pass to any instantiation should there be any checks whether the type is the one which we expect.

rgpt:
@whimboo updated PR. Please review

requirements/tests.txt
whimboo: As best yes. Looks like that this slipped through.

tests/try_scraper/test_invalid_revision.py
whimboo: Good point. Those tests we don't have yet. So if you have something in mind feel free to create those in a separate test file.

tests/try_scraper/test_invalid_revision.py
```diff
@@ -4,25 +4,16 @@
...
-                       )
+def test_invalid_parameters(httpd, tmpdir, mocker):
+    """Testing download scenarios with invalid parameters for TryScraper"""
+    query_builds_by_revision = mocker.patch('mozdownload.treeherder.Treeherder.query_builds_by_revision')
+    query_builds_by_revision.return_value = []
+    with pytest.raises(errors.NotFoundError):
+        TryScraper(destination=tmpdir,
+                   base_url=httpd.get_url(),
+                   platform='win32',
+                   revision='abc')
```
whimboo: Mind adding a final empty line to make the linter happy? Same for the other file.

requirements/tests.txt
```diff
@@ -1,7 +1,8 @@
 coveralls==1.5.1
 mock==2.0.0
-mozfile >= 1.1
-mozprocess >= 0.22
+mozfile==1.1
```
whimboo: When you use strict versions, please try to make use of the latest release which currently is 2.0.0 and is Python3 compatible. Please also check the others.

requirements/tests.txt
```diff
@@ -1,7 +1,8 @@
...
+mozprocess==0.22
```
whimboo: I don't see this package used directly anymore. Is there an indirect import from one of the other packages?

tests/try_scraper/test_invalid_revision.py
```diff
@@ -4,25 +4,16 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from mock import patch
-
+import pytest
 from mozdownload import TryScraper
```
whimboo: While `pytest` is a global import, the mozdownload ones are local. As such please separate those blocks with an empty line. Same for the other test.

requirements/tests.txt
rgpt: I could not find any use of the same.

rgpt:
@whimboo Updated PR. Please Review

requirements/tests.txt
whimboo: So I would say lets get it removed.

PR #530, opened 2019-01-28
Convert base_scraper and direct_scraper tests to Pytest

rgpt:
- Pytest ran from command line - Successful
- Pylint Checked

### DIFF
--- a/tests/base_scraper/test_base_scraper.py
+++ b/tests/base_scraper/test_base_scraper.py
@@ -19,6 +19,7 @@ def test_platform_regex(tmpdir, platform_key, platform_value):
     scraper = mozdownload.Scraper(destination=str(tmpdir), platform=platform_key)
     assert scraper.platform_regex == platform_value
 
+
 def test_standard_download(httpd, tmpdir):
     """Test standard download method"""
     filename = 'download_test.txt'
@@ -26,6 +27,7 @@ def test_standard_download(httpd, tmpdir):
     scraper = mozdownload.DirectScraper(url=test_url, destination=str(tmpdir))
     scraper.download()
     assert os.path.isfile(os.path.join(str(tmpdir), filename))
+
 
 def test_compare_download(httpd, tmpdir):
     """Compare original and downloaded file via md5 hash"""
@@ -37,31 +39,36 @@ def test_compare_download(httpd, tmpdir):
     md5_downloaded = create_md5(os.path.join(str(tmpdir), filename))
     assert md5_original == md5_downloaded
 
-def test_request_exception(httpd, tmpdir):
-    """RequestException"""
-    test_url1 = urljoin(httpd.get_url(), 'does_not_exist.html')
-    scraper1 = mozdownload.DirectScraper(url=test_url1, destination=str(tmpdir))
+
+def test_url_not_found(httpd, tmpdir):
+    """Not Found Exception"""
+    test_url = urljoin(httpd.get_url(), 'does_not_exist.html')
+    scraper = mozdownload.DirectScraper(url=test_url, destination=str(tmpdir))
     with pytest.raises(requests.exceptions.RequestException):
-        scraper1.download()
+        scraper.download()
+
 
 def test_retry_attempts(httpd, tmpdir):
     """Covering retry attempts"""
-    test_url2 = urljoin(httpd.get_url(), 'does_not_exist.html')
-    scraper2 = mozdownload.DirectScraper(url=test_url2,
+    test_url = urljoin(httpd.get_url(), 'does_not_exist.html')
+    scraper = mozdownload.DirectScraper(url=test_url,
                                          destination=str(tmpdir),
                                          retry_attempts=3,
-                                         retry_delay=1.0)
+                                         retry_delay=0.1)
     with pytest.raises(requests.exceptions.RequestException):
-        scraper2.download()
+        scraper.download()
+
 
 @pytest.mark.parametrize('attr', ['binary', 'binary_regex', 'path_regex'])
-def test_notimplementedexceptions(tmpdir, attr):
+def test_not_implemented(tmpdir, attr):
     """test implementations available"""
     scraper = mozdownload.Scraper(destination=str(tmpdir))
     with pytest.raises(errors.NotImplementedError):
         getattr(scraper, attr)
+
     with pytest.raises(errors.NotImplementedError):
         scraper.build_filename('invalid binary')
+
 
 @pytest.mark.skip(reason="Permanent failure due to mozqa.com not available anymore (#492)")
 def test_invalid_authentication(tmpdir):
@@ -70,6 +77,7 @@ def test_invalid_authentication(tmpdir):
     scraper = mozdownload.DirectScraper(destination=str(tmpdir), url=basic_auth_url)
     with pytest.raises(requests.exceptions.HTTPError):
         scraper.download()
+
 
 @pytest.mark.skip(reason="Permanent failure due to mozqa.com not available anymore (#492)")
 def test_valid_authentication(tmpdir):
@@ -84,7 +92,8 @@ def test_valid_authentication(tmpdir):
     scraper.download()
     assert os.path.isfile(os.path.join(str(tmpdir), 'mozqa.com'))
 
-def test_destination_isdirectory(httpd, tmpdir):
+
+def test_destination_as_directory(httpd, tmpdir):
     """destination is directory"""
     filename = 'download_test.txt'
     test_url = urljoin(httpd.get_url(), filename)
@@ -92,7 +101,7 @@ def test_destination_isdirectory(httpd, tmpdir):
     assert scraper.filename == os.path.join(str(tmpdir), filename)
 
 
-def test_destination_hasdirectory(httpd, tmpdir):
+def test_destination_has_directory(httpd, tmpdir):
     """destination has directory path with filename"""
     filename = 'download_test.txt'
     test_url = urljoin(httpd.get_url(), filename)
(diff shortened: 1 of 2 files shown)


rgpt:
@whimboo Added PR. Please Review

tests/base_scraper/test_base_scraper.py
```diff
@@ -5,133 +5,115 @@
...
+    scraper.download()
+    assert os.path.isfile(os.path.join(tmpdir, 'mozqa.com'))
+
+def test_destination(httpd, tmpdir):
+    """Test for various destination scenarios"""
+    tmpdir = str(tmpdir)
+    filename = 'download_test.txt'
+    test_url = urljoin(httpd.get_url(), filename)
+
+    # destination is directory
```
whimboo: I would split all the different sections into separate tests.

tests/direct_scraper/test_direct_scraper.py
```diff
@@ -5,29 +5,22 @@
...
-        self.assertTrue(os.path.isfile(os.path.join(self.temp_dir,
-                                                    scraper.filename)))
+    tmpdir = str(tmpdir)
+    filename = 'download_test.txt'
+    test_url = urljoin(httpd.get_url(), filename)
+    scraper = DirectScraper(url=test_url, destination=tmpdir)
+    assert scraper.url == test_url
+    assert scraper.filename == os.path.join(tmpdir, filename)
+    with pytest.raises(errors.NotImplementedError):
+        getattr(scraper, attr)
```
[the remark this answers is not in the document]
rgpt: If we move out this exception to a separate test then dont you think we need to extract scraper object which will be by following same procedure in two different functions for `assert os.path.isfile(os.path.join(str(tmpdir), scraper.filename))` and for `    with pytest.raises(errors.NotImplementedError):
        getattr(scraper, attr)` ?
we could run pytest.raises in a loop for it as a variable instead of parametrizing it.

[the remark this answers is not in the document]
rgpt: I agree separating exception would be better as decorator would run multiple times for the parametrized data in which part of code will be running again. I will make the asked changes on this.

tests/base_scraper/test_base_scraper.py
```diff
@@ -5,133 +5,120 @@
...
+    """Compare original and downloaded file via md5 hash"""
+    filename = 'download_test.txt'
+    test_url = urljoin(httpd.get_url(), filename)
+    scraper = mozdownload.DirectScraper(url=test_url, destination=str(tmpdir))
+    scraper.download()
+    md5_original = create_md5(os.path.join(httpd.router.doc_root, filename))
+    md5_downloaded = create_md5(os.path.join(str(tmpdir), filename))
+    assert md5_original == md5_downloaded
+
+def test_request_exception(httpd, tmpdir):
```
whimboo: I would call the test `test_url_not_found`.

tests/base_scraper/test_base_scraper.py
```diff
@@ -5,133 +5,120 @@
...
+    """RequestException"""
+    test_url1 = urljoin(httpd.get_url(), 'does_not_exist.html')
+    scraper1 = mozdownload.DirectScraper(url=test_url1, destination=str(tmpdir))
```
whimboo: Just `scraper` please. This also applies to the following tests.

tests/base_scraper/test_base_scraper.py
```diff
@@ -5,133 +5,120 @@
...
+    with pytest.raises(requests.exceptions.RequestException):
```
whimboo: Actually I wonder why we don't raise a `NotFoundError` here, which would be more appropriate. Mind filing a new issue about that?

tests/base_scraper/test_base_scraper.py
```diff
@@ -5,133 +5,120 @@
...
+        scraper1.download()
+
+def test_retry_attempts(httpd, tmpdir):
+    """Covering retry attempts"""
+    test_url2 = urljoin(httpd.get_url(), 'does_not_exist.html')
+    scraper2 = mozdownload.DirectScraper(url=test_url2,
+                                         destination=str(tmpdir),
+                                         retry_attempts=3,
+                                         retry_delay=1.0)
```
whimboo: Please remove or reduce the delay to something around 100ms. It won't make a difference anyway.

tests/base_scraper/test_base_scraper.py
```diff
@@ -5,133 +5,120 @@
...
+    password = 'mozilla'
+    basic_auth_url = 'http://mozqa.com/data/mozqa.com/http_auth/basic/'
+    scraper = mozdownload.DirectScraper(destination=str(tmpdir),
+                                        url=basic_auth_url,
+                                        username=username,
+                                        password=password)
+    scraper.download()
+    assert os.path.isfile(os.path.join(str(tmpdir), 'mozqa.com'))
+
+def test_destination_isdirectory(httpd, tmpdir):
```
whimboo: Can we rename it to `destination_as_directory`?

tests/base_scraper/test_base_scraper.py
```diff
@@ -5,133 +5,120 @@
...
+    """destination is directory"""
+    filename = 'download_test.txt'
+    test_url = urljoin(httpd.get_url(), filename)
+    scraper = mozdownload.DirectScraper(url=test_url, destination=str(tmpdir))
+    assert scraper.filename == os.path.join(str(tmpdir), filename)
+
+
+def test_destination_hasdirectory(httpd, tmpdir):
```
whimboo: You stripped an important part of the former comment. This is a directory with a filename. Also having a good method name would make the doc comment obsolete. So best would be to use the content of the former comments as method name.

tests/base_scraper/test_base_scraper.py
```diff
@@ -5,133 +5,120 @@
...
+    """destination has directory path with filename"""
+    filename = 'download_test.txt'
+    test_url = urljoin(httpd.get_url(), filename)
+    destination = os.path.join(str(tmpdir), filename)
+    scraper = mozdownload.DirectScraper(url=test_url, destination=destination)
+    assert scraper.filename == destination
+
```
whimboo: note it needs two empty lines between test methods.

tests/base_scraper/test_base_scraper.py
```diff
@@ -5,133 +5,120 @@
...
+def test_destination_hasfile(httpd):
```
whimboo: `destination_as_filename` please.

tests/base_scraper/test_base_scraper.py
```diff
@@ -5,133 +5,120 @@
...
+    """destination only has filename"""
+    filename = 'download_test.txt'
+    test_url = urljoin(httpd.get_url(), filename)
+    scraper = mozdownload.DirectScraper(url=test_url, destination=filename)
+    assert scraper.filename == os.path.abspath(filename)
+
+def test_destination_doesnotexist(httpd, tmpdir):
```
whimboo: Please separate each word with an underscore.

tests/direct_scraper/test_direct_scraper.py
```diff
@@ -5,29 +5,29 @@
...
-
-        scraper.download()
-        self.assertTrue(os.path.isfile(os.path.join(self.temp_dir,
-                                                    scraper.filename)))
+    filename = 'download_test.txt'
+    test_url = urljoin(httpd.get_url(), filename)
+    scraper = DirectScraper(url=test_url, destination=str(tmpdir))
+    assert scraper.url == test_url
+
+    assert scraper.filename == os.path.join(str(tmpdir), filename)
```
whimboo: nit: there was no empty line before.

whimboo:
> @whimboo PR Updated. Please Review.
> CI/CD - failed due to test_thunderbird - `Not Found for url: https://archive.mozilla.org/pub/thunderbird/tinderbox-builds/comm-central-win32-debug/'`

Hm, that's a remote test? Maybe when you updated the PR the test will pass. Otherwise we will have to skip this test again because building Thunderbird seems to fail since the version bump.

tests/base_scraper/test_base_scraper.py
rgpt: I was worried it would make function names too big. But will follow the exact pep standards keeping the function names concise.

tests/base_scraper/test_base_scraper.py
rgpt: I kept this specifically in my mind but i guess i missed it :( . Will surely take care of it.

tests/base_scraper/test_base_scraper.py
rgpt: It was added as a doc string in the function itself. I will rename the function as you asked. Please let me know if i missed anything.

tests/base_scraper/test_base_scraper.py
rgpt: I tried using errors.NotFoundError from mozdownload.errors but it resulted in failing of the test. I am guessing it is due to exception raised by scraper.download - https://github.com/mozilla/mozdownload/blob/master/mozdownload/scraper.py#L319. This inference could be wrong.

rgpt:
@whimboo  Updated PR Please Check. Also please take a look at https://github.com/mozilla/mozdownload/pull/530#discussion_r253270960.

tests/base_scraper/test_base_scraper.py
rgpt: As the issue is created separately for this and dedicatedly in milestone so this PR handles what was asked. I wish to take issue 534 if you guide me through it based on the comments mentioned above.

whimboo: Sure. We can follow-up on that other issue. I'm resolving this one now.

rgpt:
@whimboo PR updated. Please review

tests/base_scraper/test_base_scraper.py
whimboo: That looks better. Thanks!

PR #554, opened 2019-04-25
Skip remote tests for Thunderbird tinderbox builds

andreheringer:
This should be enough. I'm opening this PR from the issue branch I've created as I've interpreted the instructions from the CONTRIBUTING.md file.

### DIFF
--- a/tests/remote/test_thunderbird.py
+++ b/tests/remote/test_thunderbird.py
@@ -88,9 +88,9 @@ def test_daily_scraper(tmpdir, args):
     {'application': 'thunderbird', 'branch': 'comm-central', 'platform': 'win32', 'locale': 'de'},
     {'application': 'thunderbird', 'branch': 'comm-central', 'platform': 'win32', 'extension': 'txt'},
 ])
+@pytest.mark.xfail(run=False, reasson="tinderbox builds not available in the archive")
 def test_tinderbox_scraper(tmpdir, args):
     """Test tinderbox scraper against the remote server."""
-    pytest.skip("Thunderbird tinderbox not been updated to the archive")
     mozdownload.TinderboxScraper(destination=tmpdir, **args)
 
 
--- a/tests/remote/test_thunderbird.py
+++ b/tests/remote/test_thunderbird.py
@@ -85,7 +85,7 @@ def test_daily_scraper(tmpdir, args):
     {'application': 'thunderbird', 'branch': 'comm-central', 'platform': 'mac'},
     # Currently no debug builds are shipped for comm-central
     # {'application': 'thunderbird', 'branch': 'comm-central', 'platform': 'win32', 'debug_build': True},
-    # Tindebox builds are not available at the moment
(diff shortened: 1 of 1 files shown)


andreheringer:
PR related to #550

whimboo:
Thank you for the patch and it already looks promising. All the tests are passing now. But, could you please check how to correctly skip a test? See https://docs.pytest.org/en/latest/skipping.html. Maybe we should even better use `@pytest.mark.xfail` with strict as True, as I have just seen.

andreheringer:
I see your point. I've had used the implementation I was most aware off, but the `@pytest.mark.xfail` decorator would be a better fit. Strict should be True as well. Still I think we should keep the reason parameter in the implementation so, in the near future, we'll know why the tests were skipped.

tests/remote/test_thunderbird.py
```diff
@@ -88,9 +88,9 @@ def test_daily_scraper(tmpdir, args):
     {'application': 'thunderbird', 'branch': 'comm-central', 'platform': 'win32', 'locale': 'de'},
     {'application': 'thunderbird', 'branch': 'comm-central', 'platform': 'win32', 'extension': 'txt'},
 ])
+@pytest.mark.xfail(run=False, reasson="tinderbox builds not available in the archive")
```
whimboo: I would still prefer the `strict` argument here, given that once we have the missing feature implemented the test will unexpectedly pass. Otherwise we might miss to re-enable the test. Also please fix `s/reasson/reason/`

whimboo:
Hi @andreheringer. Would you have the time to implement the latest requests? That would be fantastic. Thanks!

andreheringer:
Hi @whimboo, sorry for the delay, I've implemented the requests but using this implementation there is one test that still fails :( I was trying to come up with a elegant solution that would make all tests either Xfail or skip but could not think of any.

whimboo:
Strange. Do you know what that exactly fails? Is there more output locally? It would be good to know how this test actually invokes the tinderbox scraper. When I check on archive.mozilla.org there is no mozilla-central folder nor l10n specific folder. As such I would be interested in the URL the scraper creates.

whimboo:
I would also be ok in just commenting this specfic test out of the arguments, and add a comment similar to the missing debug builds.

whimboo:
No worries. I don't think it makes that much sense to investigate that now. Lets simply get the changes landed so that we have passing tests again.

I've made a small update to the description of the commented out test to better explain why we did that. I will merge your PR in a moment.

Again, thanks a lot for your contribution.  I hope you enjoyed it, and I'm happy to guide you further in case you want to learn more by working on other issues.

andreheringer:
@whimboo Thank you so much for your time and patience guiding me throughout the process. I would like to be assigned to other issues and help whenever possible.

whimboo:
@andreheringer that is great. Maybe you want to look at issue #247? It would be a good next one for you. Thanks.

PR #564, opened 2019-06-10
Solution for Invalid Build number issue

0xDivyanshu:
I have edited few lines to provide support in case build is not available. Now it would raise an error if a incorrect non-exsistent build is used

### DIFF
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -696,7 +696,12 @@ class ReleaseCandidateScraper(ReleaseScraper):
         self.show_matching_builds(parser.entries)
         self.builds = parser.entries
         self.build_index = len(parser.entries) - 1
-
+        
+        self.tmp = "build"+str(self.build_number)
+        if self.tmp not in self.builds:
+          message = 'Selected build is not available.'
+          raise errors.NotSupportedError(message)
+          
         if self.build_number and \
                 ('build%s' % self.build_number) in self.builds:
             self.builds = ['build%s' % self.build_number]
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -696,13 +696,6 @@ class ReleaseCandidateScraper(ReleaseScraper):
         self.show_matching_builds(parser.entries)
         self.builds = parser.entries
         self.build_index = len(parser.entries) - 1
-        
-        if self.build_number:
-          self.tmp = "build"+str(self.build_number)
-          if self.tmp not in self.builds:
-            message = 'Selected build is not available.'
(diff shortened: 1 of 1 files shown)


mozdownload/scraper.py
```diff
@@ -696,7 +696,14 @@ def get_build_info(self):
         self.show_matching_builds(parser.entries)
         self.builds = parser.entries
         self.build_index = len(parser.entries) - 1
-
+        
+        if self.build_number:
+          self.tmp = "build"+str(self.build_number)
+          if self.tmp not in self.builds:
+            message = 'Selected build is not available.'
+            sys.logger.info(message)
```
whimboo: When you check the log files of the failed tests on Travis and AppVeyor you can see that `sys.logger` is not available.

mozdownload/scraper.py
```diff
@@ -696,7 +696,14 @@ def get_build_info(self):
         self.show_matching_builds(parser.entries)
         self.builds = parser.entries
         self.build_index = len(parser.entries) - 1
-
+        
+        if self.build_number:
```
whimboo: You don't want to add this code here but integrate it in the condition right below.

mozdownload/scraper.py
```diff
@@ -696,7 +696,14 @@ def get_build_info(self):
...
+          self.tmp = "build"+str(self.build_number)
+          if self.tmp not in self.builds:
+            message = 'Selected build is not available.'
+            sys.logger.info(message)
+            sys.exit()
```
whimboo: Never call `sys.exit()` but raise an exception. Have a look at the ones mozdownload already defines.

Divyanshu29:
After doing the changes as suggested It still fails at one point of Error notsupported! Can u tell me the issue? It works fine on my system.!

whimboo:
@divyanshu29 can you please rebase to latest master (with the fix from PR #566 included), and push the changes again?

divyanshu29:
This is gonna fail again .... :( @whimboo soory to disturb u again but plz have a look!

whimboo:
Yes, it's failing because the tests have to be updated to accept such an exception if the specific build number is not present, eg one tests is checking `build2` which doesn't exist here: https://github.com/mozilla/mozdownload/tree/master/tests/data/firefox/candidates/23.0.1-candidates.

So pick those tests by test file and move them as best to a new test class for invalid builds.

whimboo:
Hi @divyanshu29. Are you planning to continue on this PR? Thanks.

whimboo:
Not needed anymore given that PR #607 fixed it.

Issue #543, opened 2019-03-13
Add "remote" pytest marker for remote tests and only run those tests in CI


whimboo:
With issue #468 fixed soon, we can now mark specific functions to be not run by default when testing locally, but only when checks run in CI. This specifically applies to all the remote tests in https://github.com/mozilla/mozdownload/tree/master/tests/remote. 

A [pytest marker](https://docs.pytest.org/en/latest/example/markers.html) should be added for each of those tests, and registered in the pytest.ini file.

By default those tests should not be run, but only when being executed in CI (Travis CI, AppVeyor). To get this done their appropriate configuration (yaml) files need to be updated.

ali-alsabbah:
I'd be interested in doing this if I can get a little help

whimboo:
Absolutely! Let me know when you have questions, or also join us on irc.mozilla.org in the #automation channel for direct help. Thanks!

snowmanstark:
@whimboo Can I start working on this issue if it is not assigned to anyone?

whimboo:
@snowmanstark, sorry for the late reply but I missed your question here. You could have been started, but now we already have PR #572, which adds this feature. If you want to work on something else I would be still happy to mentor you.

snowmanstark:
Thanks @whimboo for the update! Will let you know when I would need mentoring.

whimboo:
Fixed by PR #572. Thanks a lot!

PR #572, opened 2019-08-03
Add markers for remote tests

dmitrypolo:
This addresses #543.


pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
+markers =
+    ci_vendor: Only run these tests within the CI Vendor environment.
```
whimboo: Mind renaming this marker to `ci_only`? It better describes what it is used for. Also please fix-up the sentence and remove the `vendor`.

tests/conftest.py
```diff
@@ -16,3 +16,12 @@ def httpd():
     httpd.start(block=False)
     yield httpd
     httpd.stop()
+
+
+def pytest_runtest_setup(item):
+    valid_ci = os.getenv('CI', False)
```
whimboo: Mind renaming to `ci_enabled`?

pytest.ini
dmitrypolo: done

tests/conftest.py
dmitrypolo: done

dmitrypolo:
@whimboo changed per your comments

whimboo:
I fixed two small things, and pushed those commits to your branch. It looks all fine now, so I will merge it. Thanks a lot for getting it added!

PR #596, opened 2020-03-05
Replaces RequestException with NotFoundError

### LINKED ISSUE #534: NotFoundError instead of RequestException in base_scraper test
When not able to download the package it should result into NotFoundError and not Request Exception error as the test case was specifically designed for `doesnotexist` Case.
Reference -https://github.com/mozilla/mozdownload/blob/f8474674808b3ded69354f8a09da208c03691aaf/tests/base_scraper/test_base_scraper.py#L55

kaushiksk:
Fixes #534

### DIFF
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -310,13 +310,19 @@ class Scraper(object):
 
                 if log_level <= logging.INFO and content_length:
                     pbar.finish()
+
+            except requests.exceptions.RequestException:
+                if os.path.isfile(tmp_file):
+                    os.remove(tmp_file)
+                raise errors.NotFoundError
+
             except Exception:
                 if os.path.isfile(tmp_file):
                     os.remove(tmp_file)
                 raise
 
         self._retry(_download,
-                    retry_exceptions=(requests.exceptions.RequestException,
+                    retry_exceptions=(errors.NotFoundError,
                                       errors.TimeoutError))
 
         os.rename(tmp_file, self.filename)
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -311,10 +311,13 @@ class Scraper(object):
                 if log_level <= logging.INFO and content_length:
                     pbar.finish()
 
-            except requests.exceptions.HTTPError:
-                if os.path.isfile(tmp_file):
-                    os.remove(tmp_file)
-                raise errors.NotFoundError("The requested file was not found", self.url)
+            except requests.exceptions.HTTPError as ex:
+                if ex.response.status_code == 404:
+                    if os.path.isfile(tmp_file):
+                        os.remove(tmp_file)
+                    raise errors.NotFoundError("The requested file was not found", self.url)
+                else:
+                    raise
 
             except Exception:
                 if os.path.isfile(tmp_file):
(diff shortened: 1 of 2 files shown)


kaushiksk:
The travic job failed because the py36 interpreter could not be found. Probably something that needs fixing?
It says my newly introduced lines don't have code coverage, when in fact they do.

kaushiksk:
Summary of the changes:

The exact `RequestException` thrown is the `HTTPError` (specifically HTTP 404 since the file is not found). I've generalized all HTTPError to be NotFoundError for now. In case the auth related tests are activated in the future, we will need to check the exact HTTPError (401/404 etc) and raise accordingly.

kaushiksk:
Since #597 fixes the auth tests, I've updated this change to raise NotFoundError only when HTTP 404 error occus

mozdownload/scraper.py
```diff
@@ -310,13 +310,22 @@ def _download():
 
                 if log_level <= logging.INFO and content_length:
                     pbar.finish()
+
+            except requests.exceptions.HTTPError as ex:
```
whimboo: Please move the whole block into the `Exception` section below. Note that `HTTPError` is based on `Exception` so just have a type check and handle a 404 via a if condition. That way we can reduce the code duplication.

mozdownload/scraper.py
```diff
@@ -311,19 +311,15 @@ def _download():
...
                     pbar.finish()
 
-            except requests.exceptions.HTTPError as ex:
+            except Exception as ex:
                 if os.path.isfile(tmp_file):
                     os.remove(tmp_file)
-                if ex.response.status_code == 404:
-                    raise errors.NotFoundError("The requested file was not found", self.url)
+                if type(ex) == requests.exceptions.HTTPError:
+                    if ex.response.status_code == 404:
```
whimboo: Please use `is` to compare the exception types, and also combine both if conditions into a single one. Right now we won't raise any exception if it's of type `HTTPError` and if status is not 404.

kaushiksk: I have made this change

tests/base_scraper/test_base_scraper.py
```diff
@@ -52,7 +52,7 @@ def test_retry_attempts(httpd, tmpdir):
                                         destination=str(tmpdir),
                                         retry_attempts=3,
                                         retry_delay=0.1)
-    with pytest.raises(requests.exceptions.RequestException):
+    with pytest.raises(errors.NotFoundError):
```
kaushiksk, line 55: @whimboo I have updated the two tests here.

kaushiksk:
@whimboo I have already made a changes in `test_url_not_found` and `test_retry_attempts.

mozdownload/scraper.py
```diff
@@ -310,13 +310,18 @@ def _download():
...
                     pbar.finish()
-            except Exception:
+
+            except Exception as ex:
                 if os.path.isfile(tmp_file):
                     os.remove(tmp_file)
-                raise
+                if type(ex) is requests.exceptions.HTTPError and \
+                        ex.response.status_code == 404:
+                    raise errors.NotFoundError("The requested file was not found", self.url)
```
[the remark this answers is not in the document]
whimboo: I actually fixed it myself and pushed it to your branch.

whimboo:
@kaushiksk thanks for your patch!

kaushiksk:
I only just saw your comment. Thanks for making that change! :D Cheers!

Issue #600, opened 2020-03-19
Travis CI jobs are failing on MacOS with Python 3.6 not found


whimboo:
For example:
https://travis-ci.org/github/mozilla/mozdownload/jobs/659485785?utm_medium=notification&utm_source=github_status

I assume that this is happening because for MacOS the .travis.yml doesn't contain the `python: 3.6` line. Lets see...

PR #601, opened 2020-03-19
Fix Python 3 job on MacOS (#600)

whimboo:
Fixes issue #600.


whimboo:
Travis CI is actually pretty bad when it comes to Python support on MacOS and especially a specific one. Instead of having failing jobs we should just request the 3.x version as installed for the given osx image.

PR #608, opened 2020-06-14
Improved installation section of README

ashellunts:
Removed obsolete link to wiki (does it still exist somewhere?) . Also added virtualenv commands because otherwise 'python setup.py develop' fails.

For me as a beginner it was confusing why 'python setup.py develop' fails when I followed instructions. With this change, installation process for a new contributer should go smoother :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,15 +20,14 @@ pip install mozdownload
 Otherwise follow the steps below to setup a development environment. It is
 recommended that [virtualenv](http://virtualenv.readthedocs.org/en/latest/installation.html)
 and [virtualenvwrapper](http://virtualenvwrapper.readthedocs.org/en/latest/)
-be used in conjunction with mozdownload. Start by installing these. Then first fork
-our repository into your own github account, and run:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,11 +25,14 @@ our repository into your own github account, and run:
 ```bash
 git clone https://github.com/%your_account%/mozdownload.git
 cd mozdownload
-python setup.py develop
+virtualenv venv # configures virtual environment in venv folder
+source venv/bin/activate # activates virtual enviroment
```
whimboo, line 29: Instead of adding the steps here, I would suggest to improve the text above to:

`Start by installing these, and [creating a virtualenv for the project](https://docs.python-guide.org/dev/virtualenvs/#lower-level-virtualenv)`.

This will give instructions for all the basics you have to know.

ashellunts: ok!

whimboo:
Thanks a lot for your patch. This looks fine now, and I will get it merged immediately.

Issue #609, opened 2020-07-29
Deprecation warning due to invalid escape sequences


tirkarthi:
Deprecation warnings are raised due to invalid escape sequences. This can be fixed by using raw strings or escaping the literals. pyupgrade also helps in automatic conversion : https://github.com/asottile/pyupgrade/

```
find . -iname '*.py' | grep -Ev 'vendor|example|doc|tools|sphinx' | xargs -P4 -I{} python3.8 -Wall -m py_compile {}
./setup.py:21: DeprecationWarning: invalid escape sequence \d
  return re.findall("__version__ = '([\d\.]+)'",
```

whimboo:
Thank you for the report. @tirkarthi, would you be also interested to get this issue fixed?

tanmoydas9166:
Can I fix this issue?

tirkarthi:
@tanmoydas9166 Feel free to pick it up.

tanmoydas9166:
Sir, please check my PR.

whimboo:
@tanmoydas9166 thanks for picking up this issue. I had a look at open PRs but sadly I cannot find the one from you. Did you actually create one? It's not enough to only push the changes to your own fork.

paridhi7:
Hi, I'm new to open source. Is this issue still open?

whimboo:
I haven't seen a PR being opened by @tanmoydas9166 yet. So feel free to pick-up this issue. Thanks.

sahilsandhu:
Hi, i am new to open source, is this issue resolved or it is still opened??

whimboo:
It's currently being worked on via PR #610. But maybe you want to help with https://github.com/mozilla/mozdownload/issues/247.

rishishounak:
If the issue is still open could I join in

whimboo:
If we don't hear back from @tanmoydas9166 within two days feel free to provide your own PR. Thanks!

rishishounak:
I need some guidance to get started as I am new

whimboo:
@rishishounak please check https://github.com/mozilla/mozdownload/pull/610 for the changes and my review comments. Those will be the lines of code you will have to change. Once done submit a PR. Thanks.

rishishounak:
Please whether what I have done is correct

rishishounak:
Please check whether I have made the right changes

tanmaysinghal98:
@whimboo I am a first timer, can I pick this up?

[...]

PR #610, opened 2020-09-08
Littlechange

tanmoydas9166:
Please add me to the list  @whimboo @tirkarthi

### DIFF
--- a/meet.py
+++ b/meet.py
@@ -1,0 +1,1 @@
+from selenium import
--- a/setup.py
+++ b/setup.py
@@ -18,8 +18,8 @@ def read(*parts):
 
 
 def get_version():
-    return re.findall("__version__ = '([\d\.]+)'",
-                      read('mozdownload', 'cli.py'), re.M)[0]
+	p = re.compile(r"__version__ = '(\d.+)'")
+    return p.findall(read('mozdownload', 'cli.py'), re.M)[0]
 
 
 setup(name='mozdownload',


meet.py
```diff
@@ -0,0 +1 @@
+from selenium import
```
whimboo, line 1: Why did you add this file? It shouldn't be part of this PR.

setup.py
```diff
@@ -18,8 +18,8 @@ def read(*parts):
 
 
 def get_version():
-    return re.findall("__version__ = '([\d\.]+)'",
-                      read('mozdownload', 'cli.py'), re.M)[0]
+	p = re.compile(r"__version__ = '(\d.+)'")
```
whimboo, line 21: What was basically requested is to get rid of `\d` by replacing it with `0-9`, and not changing all the code around the regex. Please get this reverted/updated.

whimboo (changes requested):
Please see my inline comments for what's remaining to do. Also please find a better wording for the commit message, which would help to later identify the changes as being done on this PR. Thanks.

whimboo:
Fixes issue #609.

setup.py
tanmoydas9166: **I was thinking of converting** 
    return re.findall("__version__ = '([\d\.]+)'",
                      read('mozdownload', 'cli.py'), re.M)[0]
**to a raw string**
    return re.findall(r"__version__ = '([\d\.]+)'",
                      read('mozdownload', 'cli.py'), re.M)[0]
Will that work?
Help appreciated.

whimboo: That should work too. Make sure to be able to reproduce the deprecation warning first with Python3. Then it makes it easier for you to verify the changes.

whimboo:
@tanmoydas9166 do you have an update for us? Thanks.

whimboo:
Closing due to inactivity.

PR #611, opened 2020-09-16
Issue609

rishishounak:
Please check I have replaced \d with 0-9 is it fine.


rishishounak:
Please tell me whether what I have done is correct

whimboo:
PR #612 looked like the right solution so I have merged the other one.

PR #612, opened 2020-09-21
Upgrade syntax of setup.py to python 3.8+

tanmaysinghal98:
This is a fix for #609 
Please take a look and let me know.

Checked with following command
```
find . -iname '*.py' | grep -Ev 'vendor|example|doc|tools|sphinx' | xargs -P4 -I{} python3.8 -Wall -m py_compile {}
```

No deprecation warning coming.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ def read(*parts):
 
 
 def get_version():
-    return re.findall("__version__ = '([\d\.]+)'",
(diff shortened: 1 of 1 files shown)


whimboo:
I'm very sorry that it took so long to review. I now had a look at and compared to the other PRs this is the right solution. Thanks a lot!

PR #634, opened 2023-01-17
Fix pass_env values cannot contain whitespace error in CI

whimboo:


whimboo:
The remaining issue is related to the removal of the `LegacyVersion` class in the `packaging` package. I'm going to fix that in a different PR.

PR #640, opened 2023-04-04
Fix packaging requirement for version 21.3.0

whimboo:
None of the tests use the packaging module but mozdownload code itself. So my patch on #636 was actually wrong and this PR fixes it.


juliandescottes (approved):
Thanks, this seems straightforward. No usage of packaging in tests, also verified that tests run fine on a fresh clone with those changes.

PR #672, opened 2024-04-28
Raise NotFoundError if no daily build can be found for latest build date

### LINKED ISSUE #664: Downloading Fenix build with invalid platform raises an index error
Using the following arguments from a non Android platform causes the code in DailyScraper to fail with an `IndexError` because there is no such build for such a platform.

> mozdownload -a fenix -t daily -p mac

```
  File "/Users/henrik/code/mozdownload/venv/bin/mozdownload", line 33, in <module>
    sys.exit(load_entry_point('mozdownload', 'console_scripts', 'mozdownload')())
  File "/Users/henrik/code/mozdownload/mozdownload/cli.py", line 169, in cli
  File "/Users/henrik/code/mozdownload/mozdownload/factory.py", line 126, in __init__
    scraper_types[scraper_type].__init__(self, **kwargs)
  File "/Users/henrik/code/mozdownload/mozdownload/scraper.py", line 383, in __init__
  File "/Users/henrik/code/mozdownload/mozdownload/scraper.py", line 165, in __init__
    self._retry_check_404(self.get_build_info)
  File "/Users/henrik/code/mozdownload/mozdownload/scraper.py", line 180, in _retry_check_404
    self._retry(func, **retry_kwargs)
  File "/Users/henrik/code/mozdownload/mozdownload/scraper.py", line 171, in _retry
    return redo.retry(func, **retry_kwargs)
  File "/Users/henrik/code/mozdownload/venv/lib/python3.10/site-packages/redo-2.0.4-py3.10.egg/redo/__init__.py", line 170, in retry
    return action(*args, **kwargs)
  File "/Users/henrik/code/mozdownload/mozdownload/scraper.py", line 434, in get_build_info
    self.date = self.get_latest_build_date()
  File "/Users/henrik/code/mozdownload/mozdownload/scraper.py", line 456, in get_latest_build_date
    date = ''.join(parser.entries[-1].split('-')[:6])
IndexError: list index out of range
```

piri-p:
Partly fixes #664. @whimboo

### DIFF
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -452,7 +452,7 @@ class DailyScraper(Scraper):
             parser = self._create_directory_parser(url)
             parser.entries = parser.filter(r'.*%s' % self.platform_regex)
             if not parser.entries:
-                raise errors.NotSupportedError('Selected platform not available')
+                raise errors.NotSupportedError('No builds have been found')
             parser.entries.sort()
 
             date = ''.join(parser.entries[-1].split('-')[:6])
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -452,7 +452,7 @@ class DailyScraper(Scraper):
             parser = self._create_directory_parser(url)
             parser.entries = parser.filter(r'.*%s' % self.platform_regex)
             if not parser.entries:
-                raise errors.NotSupportedError('No builds have been found')
+                raise errors.NotFoundError('No builds have been found')
             parser.entries.sort()
 
             date = ''.join(parser.entries[-1].split('-')[:6])
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -452,7 +452,7 @@ class DailyScraper(Scraper):
             parser = self._create_directory_parser(url)
             parser.entries = parser.filter(r'.*%s' % self.platform_regex)
             if not parser.entries:
-                raise errors.NotFoundError('No builds have been found')
+                raise errors.NotFoundError('No builds have been found', url)
             parser.entries.sort()
 
             date = ''.join(parser.entries[-1].split('-')[:6])
--- a/mozdownload/scraper.py
+++ b/mozdownload/scraper.py
@@ -451,6 +451,8 @@ class DailyScraper(Scraper):
                           months.entries[-1] + '/')
             parser = self._create_directory_parser(url)
             parser.entries = parser.filter(r'.*%s' % self.platform_regex)
+            if not parser.entries:
+                raise errors.NotSupportedError('Selected platform not available')
             parser.entries.sort()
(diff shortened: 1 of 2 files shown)


mozdownload/scraper.py
```diff
@@ -451,6 +451,8 @@ def get_latest_build_date(self):
                           months.entries[-1] + '/')
             parser = self._create_directory_parser(url)
             parser.entries = parser.filter(r'.*%s' % self.platform_regex)
+            if not parser.entries:
+                raise errors.NotSupportedError('Selected platform not available')
```
whimboo, line 455: Given that it is not clear at this point why no build has been found - it could also be something else beside the requested platfrom - lets use a general error message like:

```suggestion
                raise errors.NotSupportedError('No builds have been found')
```

whimboo (changes requested):
Thank you for this PR @piri-p! The changes make sense and it also works fine as well. I've just a small nit (see inline).

Also I would appreciate if you could add an automated test here as well. I think that we even do not have these for other applications like Firefox and Thunderbird. So it would be good to also include tests for these as well. Thanks.

mozdownload/scraper.py
```diff
@@ -451,6 +451,8 @@ def get_latest_build_date(self):
                           months.entries[-1] + '/')
             parser = self._create_directory_parser(url)
             parser.entries = parser.filter(r'.*%s' % self.platform_regex)
+            if not parser.entries:
+                raise errors.NotSupportedError('No builds have been found')
```
whimboo, line 455: I just noticed that this should be a `NotFoundError`. Looks like it slipped through in my last review. Can you please update both the code here and the test? 

Beside that we probably should indeed have a `NotSupportedError` thrown when an unsupported platform has been specified. But this check should be done as early as possible when we know the application and which kind of platform is selected. I would suggest that you create a separate PR for that and we leave open the issue even with this PR merged.

whimboo (changes requested):

tests/daily_scraper/test_invalid_platform.py
```diff
@@ -0,0 +1,19 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+from mozdownload import DailyScraper
+import mozdownload.errors as errors
+
+@pytest.mark.parametrize('args', [
+    ({'application': 'fenix', 'platform': 'mac'}),
```
whimboo, line 13: Please note that with the follow-up PR for `NotSupportedError` this will change given that `mac` is a not supported platform. But I'm fine with landing this for now, given that it only affects Fenix builds.

whimboo:
Thank you @piri-p for the patch!

PR #676, opened 2024-05-03
Add Thunderbird daily default branch handling (#247)

### LINKED ISSUE #247: Default branch for Thunderbird should be comm-central and not mozilla-central

piri-p:
Fix #247 - Add Thunderbird daily default branch handling @whimboo
Also handle Firefox, Tinderbox

Test cases added


mozdownload/cli.py
```diff
@@ -128,7 +128,6 @@ def parse_arguments(argv):
     group = parser.add_argument_group('Daily builds', 'Extra options for daily builds.')
     group.add_argument('--branch',
                        dest='branch',
-                       default='mozilla-central',
                        metavar='BRANCH',
                        help='Name of the branch, default: "%(default)s"')
```
whimboo, line 132: I wonder if we should actually remove `%(default)s` because it's `None` anyway.

tests/daily_scraper/test_daily_scraper.py
```diff
@@ -19,46 +19,46 @@
     ({'platform': 'win32', 'branch': 'mozilla-central'},
```
whimboo, line 19: Would you mind moving this test to line 32 so that it appears after all the default tests for supported platforms? That way we have a better structure of the tests. The same applies to the other tests as well.

whimboo (changes requested):

whimboo (approved):
Thanks a lot! It looks all fine now and tests are passing as well.

PR #679, opened 2024-05-07
Don't retrieve Tinderbox directory listing when timestamp is provided (#313)

### LINKED ISSUE #313: Don't retrieve directory listing if timestamp is provided for Tinderbox builds
Retrieving a full directory listing [1] is not necessary if the timestamp is already known [2].

This is relevant for AreWeSlimYet where we often retrieve all builds for a given date range.

[1] https://github.com/mozilla/mozdownload/blob/master/mozdownload/scraper.py#L875
[2] https://github.com/mozilla/mozdownload/blob/master/mozdownload/scraper.py#L879-L883

piri-p:
Fix #313 - Don't retrieve Tinderbox directory listing when timestamp is provided @whimboo


mozdownload/scraper.py
```diff
@@ -972,41 +972,43 @@ def get_build_info_for_index(self, build_index=None):
         """Get additional information for the build at the given index."""
         url = urljoin(self.base_url, self.build_list_regex)
 
-        self.logger.info('Retrieving list of builds from %s' % url)
-        parser = self._create_directory_parser(url)
-        parser.entries = parser.filter(r'^\d+$')
+        if not self.timestamp:
```
whimboo, line 975: Would you mind moving the case with a timestamp up here? It's a personal taste but having less complex conditions including not-negative ones make the code easier to read and follow. Here it's not a problem but it would be good to follow that pattern.

whimboo (changes requested):
With the patch applied it works great and removes a lot of work from the server. No new tests need to be added given that existing ones cover it well enough and cannot easily check that the directory index is not scraped.

I've just a small nit. Once applied we can get the patch merged. 

Again thanks a lot for all the great work! And in case of questions feel free to join us in the https://chat.mozilla.org/#/room/#webdriver:mozilla.org channel. It's not directly related to this project but it's the place where we all hang out.

whimboo (approved):
Thanks!

PR #683, opened 2024-05-20
Create GitHub Action to auto-publish to PyPI

libialany:
- Configure repository secrets for user and password of PyPI authentication: ``user: ${{ secrets.pypi_user }}`` ``password: ${{ secrets.pypi_password }}``.
- Implement package uploads exclusively for tagged commits using the [gh-action-pypi-publish](https://github.com/pypa/gh-action-pypi-publish/blob/unstable/v1/README.md).
- Integrate the [isolation feature](https://packaging.python.org/en/latest/guides/modernize-setup-py-project/) by adding the pyproject.toml file.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -66,10 +66,10 @@ jobs:
           toxenv: pylama
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Set up Python ${{ matrix.python }}
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python }}
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,82 +7,47 @@ on:
 
 jobs:
   build:
-    runs-on: ${{ matrix.os }}
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
-        # linux
-        - os: ubuntu-latest
-          python: "3.12"
-          toxenv: py312
-        - os: ubuntu-latest
-          python: "3.11"
-          toxenv: py311
-        - os: ubuntu-latest
-          python: "3.10"
-          toxenv: py310
-        - os: ubuntu-latest
-          python: "3.9"
-          toxenv: py39
-        - os: ubuntu-latest
-          python: "3.8"
-          toxenv: py38
-        # windows
-        - os: windows-latest
-          python: "3.12"
-          toxenv: py312
-        - os: windows-latest
-          python: "3.11"
-          toxenv: py311
-        - os: windows-latest
-          python: "3.10"
-          toxenv: py310
-        - os: windows-latest
-          python: "3.9"
-          toxenv: py39
-        - os: windows-latest
-          python: "3.8"
-          toxenv: py38
-        # macos
-        - os: macos-latest
-          python: "3.12"
-          toxenv: py312
-        - os: macos-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,88 @@
...
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        # linux
+        - os: ubuntu-latest
```
whimboo, line 16: We actually do not have any platform specific code included. As such we do not need separate build artifacts for the release. It will be enough to provide the tar'ed source package and the wheel. That means it will be enough to run this job for Linux and for the latest supported Python version.

.github/workflows/release.yml
```diff
@@ -0,0 +1,88 @@
...
+        with:
+          python-version: ${{ matrix.python }}
+
+      - name: Install dependencies
+        run: python -m pip install -U build
+
+      - name: Build package
+        run: python -m build    
+
+      - name: Publish package to PyPI
```
whimboo, line 82: I would suggest to split the jobs into `build` and `pypi-publish` similar to the example on https://github.com/marketplace/actions/pypi-publish. Maybe you can update the related code as well to better match

.github/workflows/release.yml
```diff
@@ -0,0 +1,88 @@
...
+        if: github.repository == 'mozilla/mozdownload' && github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          user: ${{ secrets.pypi_user }}
+          repository-url: https://test.pypi.org/legacy/
+          password: ${{ secrets.pypi_password }}
```
whimboo, line 88: Validation has to happen via a token and not by using a known user. That will make it more secure and easier to update or revoke permissions.

.github/workflows/release.yml
```diff
@@ -0,0 +1,88 @@
...
+          python-version: ${{ matrix.python }}
+
+      - name: Install dependencies
+        run: python -m pip install -U build
+
+      - name: Build package
+        run: python -m build    
+
+      - name: Publish package to PyPI
+        if: github.repository == 'mozilla/mozdownload' && github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
```
whimboo, line 83: Could we change the check for `mozilla/mozdownload` to the existence of the secret token identifier in the environment?

pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
+requires = ['setuptools>=40']
```
whimboo, line 2: For what specifically is this file needed?

.github/workflows/release.yml
libialany: Thank you, I had that question.

.github/workflows/release.yml
libialany: Thank you, I'll be creating two jobs.

.github/workflows/release.yml
libialany: Thank you for the advice.

pyproject.toml
libialany: It's used to configure setuptools with [pyproject.toml](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html), but it's not necessary.

whimboo:
FYI a new PR was created for the requested changes which is #686.

Issue #669, opened 2024-04-15
Create GitHub Action to auto-publish to PyPI


whimboo:
When a new release of mozdownload is created we should use the [`pypi-publish` GitHub action](https://github.com/marketplace/actions/pypi-publish) to automatically publish the release on https://pypi.org/project/mozdownload/.

libialany:
I want to contribute, I'm familiar with GitHub Actions.

whimboo:
Re-opening because we need a follow-up PR for the [remaining issue](https://github.com/mozilla/mozdownload/pull/686#pullrequestreview-2158335172). @libialany would you be interested to work on that?

libialany:
I want. Thank you @whimboo . I have two questions: If you accept a merge request, should the GitHub Action to auto-publish run successfully(build, deploy)? Additionally, if the maintainer makes a small change, should the GitHub Action run again? In both cases, right(build, deploy)?

whimboo:
Actually I wonder if we could change the workflow to not run based on a merge, but when adding a new (release) tag to the repository. That is actually the moment when all the files are ready to get published, and therefore I use the create a new release feature on Github which doesn't have a PR.

Based on that we most likely also don't need the test jobs to run but just the build and deployment steps.

libialany:
Thank you, @whimboo!

I have an idea for handling releases. For instance, I could trigger an action only when a tag is pushed. Here’s a sample configuration:

```yaml
on:
  push:
    tags: ['1.[0-9]+.[0-9]+']
jobs:
  on-success:
 ......
```

Is this the correct way to set it up?, I noticed that the tags consistently start with "1.".

whimboo:
@libialany Running the action based on a created tag sounds fine with me. But note that we cannot limit it to a name starting with `1.`. If there is a backward incompatible change in mozdownload we would have to bump the major version. As such we could do it for each and every tag without restriction. So using `- '*'` should be fine. Otherwise we could also use a regex like `- '[0-9]+.[0-9]+.[0-9]+'` to not accidentally run the action if a non-release tag gets added - but that shouldn't happen.

whimboo:
@libialany I just stumbled over this issue again and wondered if you are still interested in this follow-up fix for the release action. Please let me know. Thanks!

libialany:
Apologies for the delayed! I'd be happy to continue with this issue.

whimboo:
Sounds great. And thanks a lot for your update!

libialany:
[...]

PR #686, opened 2024-05-22
Action to auto publish PyPI

libialany:
Fixes https://github.com/mozilla/mozdownload/issues/669.

Thank you for reviewing my latest pull request, and I've made improvements in the following ways:
1. `pypa/gh-action-pypi-publish@release/v1` accepts either a password or PyPI token. Utilize the token instead of the user's password also i change the variable name to `pypi_token`.
2. Added a secret variable for specifying the Python version. Referring to the documentation of  `actions/setup-python@v5` , I've set it to '3.11.9', aligning with the available Python versions for that GitHub Action.
3. Implemented the use of artifacts to transfer the 'dist' folder between jobs (actions/download-artifact@v4, actions/upload-artifact@v4).
4. Set condition for publishing, ensuring the last step succeeds, the secret is not empty, and the event is push and is a 
tagged commit.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,48 +6,59 @@ on:
       - master
 
 jobs:
+  check-token:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check token variable
+        env:
+          PYPI_TOKEN_VAR: ${{ secrets.pypi_token }}
+        run: |
+          if [ -z "$PYPI_TOKEN_VAR" ]; then
+            echo "Pypi token variable is empty!"
+            exit 1
+          fi
+          echo "Pypi token variable is set!"
   build:
     name: Build package
+    needs: check-token
     runs-on: ubuntu-latest
-    env:
-      PYTHON_VERSION: ${{ secrets.PYTHON_VERSION }}
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set up Python ${{ vars.PYTHON_VERSION }}
+      - name: Set up Python 3.12
         uses: actions/setup-python@v5
         with:
-          python-version: ${{ vars.PYTHON_VERSION }}
+          python-version: 3.12
 
       - name: Install dependencies
         run: python -m pip install -U build
 
-      - name: Build package
+      - name: Build a binary wheel and a source tarball
         run: python -m build
 
-      - name: Upload dist folder
+      - name: Store the distribution packages
         uses: actions/upload-artifact@v4
         with:
-          name: dist_directory
+          name: python-package-distributions
           path: dist
   
   pypi-publish:
     name: Upload release to PyPI
     runs-on: ubuntu-latest
-    needs: build
-    env:
-      TOKEN: ${{ secrets.pypi_token }}
+    environment:
+      name: pypi
+      url: https://pypi.org/p/mozdownload
+    permissions:
+      id-token: write
+    needs: [check-token, build]
     steps:
       - name: Download dist folder
         uses: actions/download-artifact@v4
-        id: download
         with:
-          name: dist_directory
+          name: python-package-distributions
           path: dist
-      - name: Publish package to PyPI 
-        if: steps.download.conclusion == 'success' && env.TOKEN != '' && github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
+      - name: Upload release to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
-          user: ${{ secrets.pypi_user }}
-          repository-url: https://test.pypi.org/legacy/
+          packages_dir: dist/
           password: ${{ secrets.pypi_token }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,7 +10,7 @@ jobs:
   deploy:
     name: Upload release to PyPI
     runs-on: ubuntu-latest
-    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    if: ${{ github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.event == 'push' }}
     environment:
       name: release
       url: https://pypi.org/p/mozdownload
@@ -32,8 +32,4 @@ jobs:
 
       - name: Upload release to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
-  on-failure:
-    runs-on: ubuntu-latest
-    if: ${{ github.event.workflow_run.conclusion == 'failure' }}
-    steps:
-      - run: echo "Test has failed"
+
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,22 +6,14 @@ on:
       - master
 
 jobs:
-  check-token:
+  deploy:
+    name: Upload release to PyPI
     runs-on: ubuntu-latest
-    steps:
-      - name: Check token variable
-        env:
-          PYPI_TOKEN_VAR: ${{ secrets.pypi_token }}
-        run: |
-          if [ -z "$PYPI_TOKEN_VAR" ]; then
-            echo "Pypi token variable is empty!"
-            exit 1
-          fi
-          echo "Pypi token variable is set!"
-  build:
-    name: Build package
-    needs: check-token
-    runs-on: ubuntu-latest
+    environment:
+      name: release
+      url: https://pypi.org/p/mozdownload
+    permissions:
+      id-token: write
     steps:
       - uses: actions/checkout@v4
 
@@ -36,29 +28,5 @@ jobs:
       - name: Build a binary wheel and a source tarball
         run: python -m build
 
-      - name: Store the distribution packages
-        uses: actions/upload-artifact@v4
-        with:
-          name: python-package-distributions
-          path: dist
-  
-  pypi-publish:
-    name: Upload release to PyPI
-    runs-on: ubuntu-latest
-    environment:
-      name: pypi
-      url: https://pypi.org/p/mozdownload
-    permissions:
-      id-token: write
-    needs: [check-token, build]
-    steps:
-      - name: Download dist folder
-        uses: actions/download-artifact@v4
-        with:
-          name: python-package-distributions
-          path: dist
       - name: Upload release to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
-        with:
-          packages_dir: dist/
-          password: ${{ secrets.pypi_token }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -66,10 +66,10 @@ jobs:
           toxenv: pylama
 
     steps:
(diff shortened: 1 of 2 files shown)


whimboo:
@libialany thanks for the update. In case another update is required after my review please to keep working on this PR and please do not create a new one again. Thanks.

.github/workflows/release.yml
```diff
@@ -0,0 +1,53 @@
...
+        id: download
+        with:
+          name: dist_directory
+          path: dist
+      - name: Publish package to PyPI 
+        if: steps.download.conclusion == 'success' && env.TOKEN != '' && github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          user: ${{ secrets.pypi_user }}
+          repository-url: https://test.pypi.org/legacy/
```
whimboo, line 52: This is not our official package but points to a test instance of pypi.

whimboo (changes requested):

.github/workflows/release.yml
```diff
@@ -0,0 +1,53 @@
...
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
+    name: Build package
+    runs-on: ubuntu-latest
+    env:
+      PYTHON_VERSION: ${{ secrets.PYTHON_VERSION }}
```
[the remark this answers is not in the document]
libialany: Thanks for your response, I will definitely do it.

.github/workflows/release.yml
```diff
@@ -0,0 +1,53 @@
...
+    env:
+      TOKEN: ${{ secrets.pypi_token }}
+    steps:
+      - name: Download dist folder
+        uses: actions/download-artifact@v4
+        id: download
+        with:
+          name: dist_directory
+          path: dist
+      - name: Publish package to PyPI
```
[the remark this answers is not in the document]
libialany: Thanks, I'll correct it.

.github/workflows/release.yml
```diff
@@ -0,0 +1,53 @@
...
+
+      - name: Set up Python ${{ vars.PYTHON_VERSION }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ vars.PYTHON_VERSION }}
+
+      - name: Install dependencies
+        run: python -m pip install -U build
+
+      - name: Build package
```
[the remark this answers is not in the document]
libialany: Thanks, I'll correct it

.github/workflows/release.yml
```diff
@@ -0,0 +1,53 @@
...
+        run: python -m build
+
+      - name: Upload dist folder
```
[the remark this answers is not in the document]
libialany: Thanks, I'll correct it

.github/workflows/release.yml
```diff
@@ -0,0 +1,53 @@
...
+        uses: actions/upload-artifact@v4
+        with:
+          name: dist_directory
```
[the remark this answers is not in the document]
libialany: Thanks, I'll correct it

.github/workflows/release.yml
libialany: Thank you. When uploading your package, you'll need to specify a repository URL. There are two main options: TestPyPi and PyPi. It's highly recommended to upload your package to TestPyPi (https://test.pypi.org/) first, before uploading to another. Note that the credentials for these two repositories are different. Do you have a custom repository/registry that you're using? I followed the guide on packaging.python.org for reference.

.github/workflows/release.yml
```diff
@@ -0,0 +1,53 @@
...
+      TOKEN: ${{ secrets.pypi_token }}
+    steps:
+      - name: Download dist folder
+        uses: actions/download-artifact@v4
+        id: download
+        with:
+          name: dist_directory
+          path: dist
+      - name: Publish package to PyPI 
+        if: steps.download.conclusion == 'success' && env.TOKEN != '' && github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
```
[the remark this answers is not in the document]
libialany: now I understand ghp-action-pypi-publish@release better and will proceed to fix the PyPi-publish job. Thank you.

.github/workflows/release.yml
```diff
@@ -0,0 +1,53 @@
...
+        uses: actions/upload-artifact@v4
+        with:
+          name: dist_directory
+          path: dist
+  
+  pypi-publish:
+    name: Upload release to PyPI
+    runs-on: ubuntu-latest
+    needs: build
+    env:
```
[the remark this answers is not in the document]
libialany: My mistake, I will correct it

libialany:
I have corrected the errors I had. Could you please review it

.github/workflows/release.yml
```diff
@@ -0,0 +1,64 @@
...
+        uses: actions/upload-artifact@v4
+        with:
+          name: python-package-distributions
+          path: dist
+  
+  pypi-publish:
+    name: Upload release to PyPI
+    runs-on: ubuntu-latest
+    environment:
+      name: pypi
```
whimboo, line 49:
```suggestion
      name: release
```

I've now [setup the PyPI project to use a trusted publisher](https://til.simonwillison.net/pypi/pypi-releases-from-github). The name of the environment is `release`. So please update the name for this particular job.

.github/workflows/release.yml
```diff
@@ -0,0 +1,64 @@
+name: Release
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  check-token:
```
whimboo, line 9: Given that we won't need a token when using a Trusted Publisher, you can remove this step. Instead I would suggest that we add an additional test job that runs the tests again to ensure that no package dependency is broken (see the link with examples in the other comment).

.github/workflows/release.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Check token variable
+        env:
+          PYPI_TOKEN_VAR: ${{ secrets.pypi_token }}
+        run: |
+          if [ -z "$PYPI_TOKEN_VAR" ]; then
+            echo "Pypi token variable is empty!"
+            exit 1
+          fi
+          echo "Pypi token variable is set!"
+  build:
```
whimboo, line 21: Given that we have to run only a single build lets combine this job and `pypi-publish` to `deploy`. That way we also don't have to cache the generated artifacts and download them again for the next job.

.github/workflows/release.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Download dist folder
+        uses: actions/download-artifact@v4
+        with:
+          name: python-package-distributions
+          path: dist
+      - name: Upload release to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          packages_dir: dist/
+          password: ${{ secrets.pypi_token }}
```
whimboo, line 64:
```suggestion

```

It should now work without these lines.

whimboo (changes requested):
Thanks a lot for the update and sorry for the delay here. I would actually propose some more changes because I've enabled a Trusted Publisher for this PyPI project. Please check my inline comments for more details.

.github/workflows/release.yml
libialany: Thank you for the advice.

.github/workflows/release.yml
libialany: Thank you advice.

.github/workflows/release.yml
libialany: I think removing the artifact quickly(5min) might be a solution. Your suggestion is more appropriate for this project.

libialany:
I have corrected the errors I had.[improved version](https://github.com/libialany/mozdownload/blob/auto-publish_PyPI/.github/workflows/release.yml). Thank you very much for the link. I didn't apply a test job because I believe the tests are fully covered with test.yml. Could you please review it?,

whimboo:
FYI I created a PR to fix the tests at https://github.com/mozilla/mozdownload/pull/692.

libialany:
Thank you!. I have improved the [release.yml](https://github.com/libialany/mozdownload/blob/auto-publish_PyPI/.github/workflows/release.yml) and attempted to integrate the `test.yml` workflow:

```yaml
workflow_run:
  workflows: [ Test ] # Name of the test.yml workflow
  types: [ completed ]    # Trigger when the test workflow completes
  branches: [ master ]    # Run in the master branch; the branch must be the default branch (if you specify a branch that isn't the default, it will never run)
```

I also added a condition to the deploy job to check if the completed test workflow was successful:

```yaml
if: ${{ github.event.workflow_run.conclusion == 'success' }}
```

I will learn more about testing in GitHub Actions and will review the documentation at [#692](https://github.com/mozilla/mozdownload/pull/692).
link resources: 
[branches](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#limiting-your-workflow-to-run-based-on-branches)
[work_on](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run)

.github/workflows/release.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - name: Build a binary wheel and a source tarball
+        run: python -m build
+
+      - name: Upload release to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+  on-failure:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'failure' }}
+    steps:
```
whimboo, line 38: A quick follow-up question (otherwise all looks fantastic) to those lines, are they really needed? When a failure happens we will still see which specific job has failed so I don't think this will deliver more helpful data. I would just remove it. Also please keep an empty line at the end of the file.

libialany: Those lines are not needed, but I just want to make the pipeline verbose, so I fixed the 
pipeline but I forgot to add a condition: `github.event.workflow_run.event == 'push'`. I think in case pull_request is not needed to upload the release.

whimboo (approved):
Thank you for the latest update! I think that we are now at a stage where we basically should land all the changes related to the `release` workflow. When the next release is ready I'll try it out and if it fails will provide a follow-up patch.

Thank you @libialany for all the hard work to make this workflow a reality! It will clearly help to better automate the release process of mozdownload, and maybe we could even add more in the future like bumping the version number, updating the changelog etc...

.github/workflows/release.yml
whimboo: So the addition of `push` was probably not correct given that we always run the `release` workflow now. Instead we should limit this workflow to PRs that are about releases only. It's very unlikely that branches except for release use only version numbers so we could check `github.ref_name` for like starting with a number?

libialany: Hello, I'll fix this

whimboo: Great. Given that this PR is closed lets follow-up on the related issue for now.

PR #690, opened 2024-06-12
Use "http" fixture per session to increase test runtime drasticallly

whimboo:
I noticed that tests that are only using data from the local HTTP server are running very slow. After investigation it turned out that the HTTP server's shutdown method takes about 0.5s to run. I filed https://github.com/web-platform-tests/wpt/issues/46710 to get this investigated.

To workaround the issue we should use the `http` fixture per session and not per test.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,4 +1,6 @@
 [pytest]
+addopts = -vv --durations=5
+log_cli = 1
 markers =
     ci_only: Only run these tests within the CI environment.
-    
+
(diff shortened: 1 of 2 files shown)


whimboo:
Compared with CI checks that were running until now the duration drops from around 3:30 minutes to just 1 minute!

Girl1981 (approved):

Girl1981 (approved):

PR #705, opened 2024-11-04
Add Python 3.13 and remove Python 3.8

whimboo:
Python 3.8 is no longer supported. Replace it with the new Python 3.13 release.


whimboo:
As it looks like we are blocked on the missing coveralls support for Python 3.13. A PR is already open and under review: https://github.com/TheKevJames/coveralls-python/pull/542

PR #709, opened 2024-12-10
Add support for tar.xz archive file formats

### LINKED ISSUE #708: Use tar.xz archives for Linux if available by default
As the following blog post explains the downloadable builds for Firefox switched from `tar.bz2` to `tar.xz`. We should update mozdownload to support this file extension and at least for newer builds prefer this extension over `tar.bz2`.

https://blog.nightly.mozilla.org/2024/11/28/announcing-faster-lighter-firefox-downloads-for-linux-with-tar-xz-packaging/

whimboo:
Fixes #708.


mozdownload/scraper.py
```diff
@@ -364,6 +365,14 @@ def _download():
 
         return self.filename
 
+    def get_file_extension(self, binary):
+        suffixes = Path(binary).suffixes
+        if suffixes[-1] in [".bz2", ".xz"]:
```
whimboo, line 370: Actually we also need to support `.tar.gz` which I missed. Also we should change the logic here given that `.tar` isn't required to be always present.

mozdownload/scraper.py
```diff
@@ -68,9 +69,9 @@
                            'android-armeabi-v7a': 'apk',
                            'android-x86': 'apk',
                            'android-x86_64': 'apk',
-                           'linux': 'tar.bz2',
-                           'linux64': 'tar.bz2',
-                           'linux-arm64': 'tar.bz2',
+                           'linux': 'tar.(bz2|xz)',
+                           'linux64': 'tar.(bz2|xz)',
+                           'linux-arm64': 'tar.(bz2|xz)',
```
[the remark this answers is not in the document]
whimboo: We usually only allow to download the binaries. If you want to download something else then you have to use the `--extension` argument. But I'm as well rewriting this code right now.

mozdownload/scraper.py
whimboo: This was kinda crappy. With the updated code we do not need this anymore.

juliandescottes (approved):
Looks good to me, thanks!

PR #712, opened 2025-02-20
fix: update locale tests from 'de' to 'fr' for improved stability

### LINKED ISSUE #693: Update remote l10n related tests for increased stability (archive.mozilla.org clean-up)
Given that a lot of old builds will be cleaned-up on archive.mozilla.org we should update our remote tests to reduce the risk of failures at some point. Affected here for now are only our checks for `de` locale builds. Those should be updated to use the `fr` (french) locale.

AnsahMohammad:
Switch German locale tests to French to prevent failures from archive cleanup. Changed the locale in `tests/remote/` from 'de' to 'fr'.

This PR closes #693


AnsahMohammad:
@whimboo  requesting your review in here.

tests/remote/test_devedition.py
```diff
@@ -26,7 +26,7 @@
      'devedition/releases/128.0b1/win32/en-US/Firefox Setup 128.0b1.exe'),
     ({'application': 'devedition', 'platform': 'win64', 'version': '128.0b1'},
      'devedition/releases/128.0b1/win64/en-US/Firefox Setup 128.0b1.exe'),
-    ({'application': 'devedition', 'platform': 'win32', 'version': '128.0b1', 'locale': 'de'},
+    ({'application': 'devedition', 'platform': 'win32', 'version': '128.0b1', 'locale': 'fr'},
      'devedition/releases/128.0b1/win32/de/Firefox Setup 128.0b1.exe'),
```
whimboo, line 30: Changing the locale in the line above will make the test fail because the expected filename will still contain `de`. You need to run the tests locally to ensure that nothing will regress. It as well applies to most of the other tests as well. I'll trigger the CI tests so that you can compare.

whimboo (changes requested):
Thank you for your first patch for this repo! Please see my inline comment for things that need to be addressed.

tests/remote/test_devedition.py
whimboo: Hi @AnsahMohammad. Are you planning to push an update for this particular PR?

AnsahMohammad: Hey, sorry for the delay, 
Yes i'll try to make a push over this weekend

whimboo:
Hi @AnsahMohammad. Do you have a chance to get this PR updated? I'm going to release a new version of mozdownload soon and it would be great to have stable tests. Thanks.

PR #716, opened 2025-03-14
Add Python 3.13 to GitHub actions

whimboo:


whimboo:
So this is blocked by coveralls not yet having Python 3.13 support (https://github.com/TheKevJames/coveralls-python/issues/549).

whimboo:
@lutien or @juliandescottes would you mind doing a review? Thanks.

Note that I do not see progress on coveralls to support Python 3.13 I moved the coverage report over to Codecov. Also the tox config is now only used for local testing.

PR #718, opened 2025-05-19
Update release action to get triggered from a new tag being created

whimboo:
Fixed https://github.com/mozilla/mozdownload/issues/669.

With the changes the release action should be triggered when a new tag is getting created.


whimboo:
@lutien or @juliandescottes maybe one of you can check the changes? Sadly there is no way to test those changes before landing. So let me know if you notice something suspicious. Thanks!

lutien (approved):

PR #742, opened 2026-07-24
Raise NotSupportedError for invalid application/platform combination

pandeylakshya207-max:
Fixes #664

## Problem
Running `mozdownload -a fenix -t daily -p mac` (or any non-Android
platform with `fenix`) crashes deep inside `DailyScraper.get_latest_build_date()`
with a cryptic `IndexError`, since fenix builds only exist for Android
platforms and the code assumes at least one matching entry will be found.

## Fix
Added `APPLICATIONS_TO_PLATFORMS`, a mapping of applications restricted
to specific platforms. `Scraper.__init__` now validates `platform` against
this mapping immediately, before any network calls are made, and raises a
clear `NotSupportedError` if the combination is invalid.

## Testing
- `mozdownload -a fenix -t daily -p mac` → clean `NotSupportedError` with
  a helpful message listing supported platforms
- `mozdownload -a fenix -t daily -p android-arm64-v8a` → still resolves
  correctly to a real build URL, confirming valid platforms are unaffected

This is my first contribution to mozdownload/Mozilla — happy to adjust
based on feedback @whimboo.

PR #743, opened 2026-08-07
Raise NotSupportedError for invalid application/platform combination

### LINKED ISSUE #664: Downloading Fenix build with invalid platform raises an index error

pandeylakshya207-max:
Fixes #664

## Problem
Running `mozdownload -a fenix -t daily -p mac` (or any non-Android
platform with `fenix`) crashes deep inside `DailyScraper.get_latest_build_date()`
with a cryptic `IndexError`, since fenix builds only exist for Android
platforms and the code assumes at least one matching entry will be found.

## Fix
Added `APPLICATIONS_TO_PLATFORMS`, a mapping of applications restricted
to specific platforms. `Scraper.__init__` now validates `platform` against
this mapping immediately, before any network calls are made, and raises a
clear `NotSupportedError` if the combination is invalid.

## Testing
- `mozdownload -a fenix -t daily -p mac` → clean `NotSupportedError` with
  a helpful message listing supported platforms
- `mozdownload -a fenix -t daily -p android-arm64-v8a` → still resolves
  correctly to a real build URL, confirming valid platforms are unaffected

Note: This is a resubmission — my fork was accidentally deleted, which
auto-closed the original PR (#742). Same fix, same testing, just
reopened here. Apologies for the noise @whimboo.


pandeylakshya207-max:
Hi @whimboo, just following up again — it's been a few weeks since this was opened. No rush at all, just want to confirm it's still on your radar. Happy to make any changes needed. Thanks!

whimboo:
> Running `mozdownload -a fenix -t daily -p mac` (or any non-Android platform with `fenix`) crashes deep inside `DailyScraper.get_latest_build_date()` with a cryptic `IndexError`, since fenix builds only exist for Android platforms and the code assumes at least one matching entry will be found.

As mentioned on the original issue this actually shouldn't be true anymore given that the `IndexError` was fixed by PR #672 about 2 years ago. With this PR we only want to follow-up with earlier checks to not even have to reach the remote servers for fetching available builds - which wouldn't even exist in those cases.

mozdownload/scraper.py
```diff
@@ -142,6 +148,13 @@ def __init__(self, destination=None, platform=None,
 
         self.platform = platform or self.detect_platform()
 
+        allowed_platforms = APPLICATIONS_TO_PLATFORMS.get(application)
+        if allowed_platforms and self.platform not in allowed_platforms:
+            raise errors.NotSupportedError(
```
whimboo: FYI: This would require related tests to be updated to now expect a `NotSupportedError` instead of `NotFoundError`. Did you [run the tests locally](https://github.com/mozilla/mozdownload#testing)? You should see the failure in the output.

mozdownload/scraper.py
```diff
@@ -89,6 +89,12 @@
                       'win32': r'win32',
                       'win64': r'win64(-x86_64)?'}
 
+# Applications restricted to a specific set of platforms
+APPLICATIONS_TO_PLATFORMS = {
+    'fenix': ('android-arm64-v8a', 'android-armeabi-v7a',
```
whimboo, line 94: Doing that for Fenix that way we should as well add restrictions for Firefox and Thunderbird for desktop only platforms.

Meanwhile I actually think that it might be better to convert the `firefox` as application name to `fenix` internally if the Android platform is chosen. Fenix is not actually a separate product but just Firefox for Android. Having that we wouldn't need such a restriction for platforms at all - at least for the moment.

But given that we cannot remove `fenix` from the supported list of applications (due to downstream consumers relying on it) we might want to keep it and do the reverse - if specified and the platform is not Android use `firefox` as application. Then we can follow-up with the firefox proposal from above in a different issue / PR.

whimboo (changes requested):
Sorry for the delay in reviewing. But now I finally got to checking your changes. Thanks for providing this PR so that we can make progress on mozdownload's stability!

You can find my comments inline. Let me know if you have further questions.

pandeylakshya207-max:
Thanks for the review, @whimboo! I've pushed updates addressing both points:

Fallback approach: fenix now silently falls back to firefox when a non-Android platform is requested, as you suggested, instead of raising NotSupportedError.

Bug found during testing: the fallback needs to happen before self.locale is determined in Scraper.__init__, since APPLICATIONS_MULTI_LOCALE membership is checked against the application value at that point. My first attempt put the fallback after locale resolution, which meant self.locale was still set to 'multi' (fenix's locale) even after application was corrected to firefox — causing the build lookup to fail, since real Firefox builds use en-US. Reordered so the platform check runs first.

pandeylakshya207-max:
Thanks for the review, @whimboo! I've pushed updates addressing both of your points:

Fallback approach: fenix now silently falls back to firefox when a non-Android platform is requested, instead of raising NotSupportedError, as you suggested.

Bug found during testing: the fallback needs to happen before self.locale is determined in Scraper.__init__, since APPLICATIONS_MULTI_LOCALE membership is checked against the application value at that point. My first attempt put the fallback after locale resolution, which meant self.locale stayed 'multi' (fenix's locale) even after application was corrected to firefox — causing the build lookup to fail, since real Firefox builds use en-US. Reordered so the platform check runs first.

Also fixed a lint failure from a leftover debug print I'd left in while diagnosing the above, and confirmed everything end-to-end with a real mozdownload -a fenix -t daily -p mac run — it now correctly resolves and downloads a Firefox mac nightly build instead of erroring.

Let me know if there's anything else you'd like changed!

pandeylakshya207-max:
Hi @whimboo, just checking in — I addressed both points from your review
(fallback to firefox instead of raising an error, and the locale-ordering
fix) and also cleaned up a lint failure. Let me know if anything else
needs changing, happy to make further adjustments. Thanks!

pandeylakshya207-max:
Hi @whimboo, just following up — wanted to check if the latest commits
addressed your review comments satisfactorily. Happy to make further
changes if needed. Thanks!

pandeylakshya207-max:
Hi @whimboo, just a gentle follow-up — it's been a couple of weeks since
I pushed the fixes for your review comments. Totally understand if
you're busy, just wanted to keep this on your radar. Let me know if
there's anything else needed. Thanks!

whimboo:
@pandeylakshya207-max I'm very sorry for the late reply here. It's still in my queue but I haven't had the time yet to re-review the patch. I should have mentioned it instead of letting you wait. I'll try to do it over the next days.

pandeylakshya207-max:
No worries at all, thanks for the update @whimboo — take your time!

